PR #2, opened 2018-11-18
Added discord embeds

taahamahdi:


kmq.js
```diff
@@ -11,34 +11,86 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
     else {
         let guess = cleanSongName(message.content);
         if (currentSong && guess === cleanSongName(currentSong)) {
             //this should be atomic
-            scoreboard[getUserIdentifier(message.author)] = (scoreboard[getUserIdentifier(message.author)] || 0) + 1;
-            message.channel.send("Correct answer was: " + currentSong + " by " + currentArtist + "\n" + JSON.stringify(scoreboard));
+            let index = -1;
+            for (let i = 0; i < scoreboard.length; i++) {
+                if (scoreboard[i].name === getUserIdentifier(message.author)) {
+                    index = i;
```
Brainicism: break out of the loop

kmq.js
```diff
@@ -11,34 +11,86 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
     }
 });
-var scoreboard = {};
+var scoreboard = [];
 var currentSong = null;
 var currentArtist = null;
+var currentSongLink = null;
 var gameInSession = false;
+
+const songInfoMessage = (message, currentSong, currentArtist, currentSongLink, isQuit) => {
```
Brainicism: method name doesn't describe what it actually does, aka send the message

kmq.js
```diff
@@ -11,34 +11,86 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
+    message.channel.send({embed: {
+        color: 15158332, // Red
```
Brainicism: extract into a constant

kmq.js
```diff
@@ -11,34 +11,86 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
+        author: {
+            name: isQuit ? null : message.author.username,
+            icon_url: isQuit ? null : message.author.avatarURL
+        },
+        title: currentSong + " - " + currentArtist,
+        description: "https://youtube.com/watch?v=" + currentSongLink,
+        image: {
+            url: "https://img.youtube.com/vi/" + currentSongLink + "/hqdefault.jpg"
```
Brainicism: for future reference, would prefer string interpolation here

kmq.js
```diff
@@ -11,34 +11,86 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
+        }
     }
     else {
         let guess = cleanSongName(message.content);
         if (currentSong && guess === cleanSongName(currentSong)) {
             //this should be atomic
-            scoreboard[getUserIdentifier(message.author)] = (scoreboard[getUserIdentifier(message.author)] || 0) + 1;
-            message.channel.send("Correct answer was: " + currentSong + " by " + currentArtist + "\n" + JSON.stringify(scoreboard));
+            let index = -1;
+            for (let i = 0; i < scoreboard.length; i++) {
```
Brainicism: Instead of having scoreboard as an array and looking up each user in O(n), you can just have an object with each user identifier as a key in O(1)

kmq.js
taahamahdi: is sendSongMessage better?

Brainicism: sure

taahamahdi:
Thanks for the feedback!

kmq.js
```diff
@@ -14,31 +15,78 @@ const db = new sqlite3.Database('./main.db', (err) => {
...
+}
+
+const sendScoreboard = (message, scoreboard) => {
+    var scoreboardArr = Object.keys(scoreboard).map(x => {
+        return {name: x, value: scoreboard[x].value}
+    })
+    message.channel.send({embed: {
+        color: RED,
+        title: "**Results**",
+        fields: scoreboardArr.sort((a, b) => { return b.value - a.value })
```
Brainicism: You can chain the map and the sort in one statement

kmq.js
```diff
@@ -56,9 +104,10 @@ const startGame = (message) => {
         let random = rows[Math.floor(Math.random() * rows.length)];
         currentSong = random.name;
         currentArtist = random.artist;
+        currentSongLink = random.youtube_link;
         console.log(currentSong);
```
Brainicism: Remove logging

kmq.js
```diff
@@ -14,31 +15,78 @@ const db = new sqlite3.Database('./main.db', (err) => {
 var scoreboard = {};
 var currentSong = null;
 var currentArtist = null;
+var currentSongLink = null;
 var gameInSession = false;
+
+const sendSongMessage = (message, currentSong, currentArtist, currentSongLink, isQuit) => {
```
Brainicism: `currentSong`, `currentArtist`, and `currentSongLink` are already global and accessible from within this function

PR #13, opened 2018-11-18
Add Helper Text

Kirakishou:
Help text comes from help_strings.json and are sent as a command

### DIFF
--- a/help_strings.json
+++ b/help_strings.json
@@ -9,7 +9,7 @@
             "name": "random",
             "description": [
                 "Bot plays a random song in VC; ",
-                "type in your guess first to get a point"
+                "type in your guess first to get a point."
             ],
             "usage": "!random",
             "arguments": []
@@ -17,7 +17,7 @@
         {
             "name": "stop",
             "description": [
-                "The bot leaves VC and reveals the answer to any ongoing games"
+                "The bot leaves VC and reveals the answer to any ongoing games."
             ],
             "usage": "!stop",
             "arguments": []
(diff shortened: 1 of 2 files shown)


help_strings.json
```diff
@@ -0,0 +1,51 @@
+{
+    "rules": [
+      "Type '!random' in chat and the bot will play ",
```
Brainicism: Is there a reason this is an array of split clauses?

kmq.js
```diff
@@ -95,6 +99,47 @@ client.on('message', message => {
     }
 });
 
+// Usage: `!help [action]` or `!help`
+const help = (message, action) => {
+    var embed_title = "";
```
Brainicism: Change all `var`s to `let`s

kmq.js
```diff
@@ -95,6 +99,47 @@ client.on('message', message => {
...
+            embed_fields.push({
+                name: argument.name,
+                value: argument.description.join("")
+            })
+        });
+    }
+    else {
+        embed_title = "KMQ Command Help"
+        embed_desc = helpMessages.rules.join("")
+        helpMessages.actions.forEach(function(action) {
```
Brainicism: Change callback to arrow function

help_strings.json
Kirakishou: readability

Brainicism: I think it'd make more logical sense to keep it as a single string. Text wrapping should be handled by the IDE

kmq.js
```diff
@@ -70,6 +73,47 @@ client.on("message", (message) => {
     }
 });
 
+// Usage: `!help [action]` or `!help`
+const help = (message, action) => {
+    let embed_title = "";
```
Brainicism: Going forward, we should use camelCase for variables but I'll just fix this manually

Issue #14, opened 2018-11-18
Fix !end command behavior


taahamahdi:
Currently, the !end command throws an error because `scoreboard` is being interpreted as an array instead of a JS object.

PR #15, opened 2018-11-18
Bot token error handling/Fixed !end [#11][#14]

taahamahdi:


kmq.js
```diff
@@ -64,9 +64,10 @@ client.on('message', message => {
             startGame(message);
         }
         else if (command.action === "end") {
-            if (scoreboard.length > 0) {
-                disconnectVoiceConnection(message);
-                message.channel.send(`${scoreboard[0].name} wins!`);
+            if (Object.keys(scoreboard).length) {
+                let voiceConnection = client.voiceConnections.get(message.guild.id);
```
Brainicism: Why use this instead of `disconnectVoiceConnection`?

kmq.js
```diff
@@ -153,4 +154,10 @@ const getUserIdentifier = (user) => {
     return `${user.username}#${user.discriminator}`
 }
 
-client.login(config.bot_token);
+if (!config.bot_token) {
+    console.error("No bot token set. Please update config.json!")
```
Brainicism: Terminate the program here

kmq.js
taahamahdi: In its current state `disconnectVoiceConnection` will also run `message.channel.send("No VC to connect to. Please issue the command again when you are in a voice channel.");`, which we don't want to send out in the case the user has left the voice channel and issues `!end`. I have this order of execution fixed in my changes for #8.

Brainicism: How does it make sense to tell the user to join a voice channel in the function that disconnects them from a voice channel?

taahamahdi: That code has been in there since the initial commit, I've moved it into the `else if (command.action === "random")` part. I can commit those changes as well, but note they also affect #8.

Brainicism: Should definitely not be in there imo

taahamahdi: I agree

Issue #4, opened 2018-11-18
Scoreboard should be unique to each server


Brainicism:
Currently global. Some other state variables also have the same problem

PR #22, opened 2018-11-22
Added back game resets; print all winners [#21]

taahamahdi:
Removing the repeated lines of code caused more issues than it solved, so I added back the original statements. In addition, I added a loop with O(n) efficiency to print all winners in case of a tie.


kmq.js
```diff
@@ -45,8 +45,18 @@ client.on("message", (message) => {
...
             if (Object.keys(scoreboard).length) {
                 disconnectVoiceConnection(message);
-                message.channel.send(`${Object.keys(scoreboard)[0]} wins!`);
+                let sortedScoreboard = Object.keys(scoreboard).map(x => {
+                    return { name: x, value: scoreboard[x].value }
+                }).sort((a, b) => { return b.value - a.value })
+                for (let i = 0; i < sortedScoreboard.length; i++) {
+                    if (sortedScoreboard[i] === sortedScoreboard[0]) {
+                        // In case of a tie
+                        message.channel.send(`${sortedScoreboard[i].name} wins!`);
```
Brainicism: Should concatenate winners instead of sending multiple messages.

taahamahdi:
Added changes to solve #4 as well.

gamesession.js
```diff
@@ -1,19 +1,42 @@
...
-    #scoreboard;
+const Scoreboard = require("./scoreboard.js");
 
+module.exports = class GameSession {
     constructor() {
-        currentSong = null;
-        currentArtist = null;
-        currentSongLink = null;
-        gameInSession = null;
+        this.song = null;
```
Brainicism: Should these be private?

kmq.js
```diff
@@ -127,53 +116,50 @@ const help = (message, action) => {
...
+    let gameSession = gameSessions[message.guild.id];
+
+    if (gameSession.inSession) {
         message.channel.send("Game already in session.");
         return;
     }
-    gameInSession = true;
-    let query = `SELECT videos.youtube_link as youtubeLink, videos.name, DATE(videos.publish_date) as date, artists.name as artist, videos.video_type as video_type, videos.dead as dead FROM videos INNER JOIN artists on videos.artistID = artists.id WHERE gender = "female" AND video_type = "main" AND dead = "n" ORDER BY views DESC LIMIT 500`;
+
+    let query = `SELECT videos.youtube_link as youtube_link, videos.name, DATE(videos.publish_date) as date, artists.name as artist, videos.video_type as video_type, videos.dead as dead FROM videos INNER JOIN artists on videos.artistID = artists.id WHERE gender = "female" AND video_type = "main" AND dead = "n" ORDER BY views DESC LIMIT 500`;
```
Brainicism: Let's keep this as camelcase

kmq.js
```diff
@@ -127,53 +116,50 @@ const help = (message, action) => {
...
-const sendSongMessage = (message, isQuit) => {
+const sendSongMessage = (message, isForfeit) => {
+    let gameSession = gameSessions[message.guild.id];
     message.channel.send({
         embed: {
             color: RED,
             author: {
-                name: isQuit ? null : message.author.username,
-                icon_url: isQuit ? null : message.author.avatarURL
+                name: isForfeit ? null : message.author.username,
```
Brainicism: !end isn't really forfeitng

scoreboard.js
```diff
@@ -1,16 +1,81 @@
...
+        this._firstPlace = [];
+        // Both players and firstPlace are arrays of Player objects
+    }
+
+    getWinner() {
+        let winnerStr = "";
+        let isTie = (this._firstPlace.length > 1);
+        for (let i = 0; i < this._firstPlace.length; i++) {
+            if (this._firstPlace.length == 1) {
+                winnerStr = this._firstPlace[i].getName() + " ";
```
Brainicism: Can't you break out of the loop here

scoreboard.js
```diff
@@ -1,16 +1,81 @@
...
+            }
+            else if (this._firstPlace.length - i == 1) {
```
Brainicism: This would be more readable as `i == this._firstPlace.length - 1`

scoreboard.js
```diff
@@ -1,16 +1,81 @@
...
+                // Last entry -- append just the username
+                winnerStr += this._firstPlace[i].getName() + " ";
+            }
+            else if (this._firstPlace.length - i == 2) {
```
Brainicism: Same here

scoreboard.js
Brainicism: In this case it isn't even a tie so you shouldn't need to enter this loop, just return the string outside of the loop.

scoreboard.js
```diff
@@ -1,16 +1,81 @@
-class Scoreboard {
+const Player = require("./player.js")
+
+module.exports = class Scoreboard {
+    constructor() {
+        // players stores each entry as:
+        // {name: "username", value: SCORE_VAL }
+        this._players = [];
```
Brainicism: Why is this an array instead of a hash

scoreboard.js
```diff
@@ -1,16 +1,81 @@
...
+        }
         if (index == -1) {
-            players.push(new Player(winner));
+            this._players.push(new Player(winner));
+            this.updateWinner();
         }
         else {
-            players[index].incrementScore();
+            this._players[index].incrementScore();
+            this.sortScoreboard();
```
Brainicism: We don't need to sort the scoreboard if you keep track of the highest score and list of people in first place.

player.js
```diff
@@ -0,0 +1,18 @@
+module.exports = class Player {
```
Brainicism: Probably want to keep track of their user_id here too

gamesession.js
taahamahdi: Yup, thanks.

kmq.js
taahamahdi: `isQuit`/`isForfeit` is only true when the user calls !stop, not !end. We don't actually call this function when a game is in progress and a user calls !end, I'll fix that

player.js
taahamahdi: Since we're storing the user_id as a key for each player, this comment may no longer apply.

gamesession.js
```diff
@@ -0,0 +1,46 @@
+const Scoreboard = require("./scoreboard.js");
+
+module.exports = class GameSession {
+    constructor() {
+        this._song = null;
+        this._artist = null;
+        this._link = null;
+        this._inSession = false;
+        this.scoreboard = new Scoreboard();
```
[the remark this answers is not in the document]
taahamahdi: We need to access its member functions, and I think it would be counter-intuitive to create more interface functions in GameSession when we can just use Scoreboard's interface functions directly.

scoreboard.js
```diff
@@ -0,0 +1,67 @@
...
+        let winnerStr = "";
+
+        if (this._firstPlace.length == 1) {
+            return this._firstPlace[0].getName() + " wins!";
+        }
+
+        for (let i = 0; i < this._firstPlace.length; i++) {
+            if (i == this._firstPlace.length - 1) {
+                // Last entry -- append just the username
+                winnerStr += this._firstPlace[i].getName() + " ";
```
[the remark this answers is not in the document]
taahamahdi: Great point, thanks.

scoreboard.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if (this._players[winnerID].getScore() == this._highestScore) {
+            // If user is tied for first, add them to the first place array
+            this._firstPlace.push(this._players[winnerID]);
+        }
+        else if (this._players[winnerID].getScore() > this._highestScore) {
+            // If user is first, reset first place array and add them
+            this._highestScore = this._players[winnerID].getScore();
+            this._firstPlace = [];
```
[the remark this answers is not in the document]
taahamahdi: Ah, you're right!

PR #31, opened 2018-12-10
User can choose artist genders [#30]

taahamahdi:


game_session.js
```diff
@@ -1,6 +1,9 @@
 const Scoreboard = require("./scoreboard.js");
 const BEGINNING_SEARCH_YEAR = 2008;
-const DEFAULT_OPTIONS = { beginningYear: BEGINNING_SEARCH_YEAR }
+const MALE = "male";
+const FEMALE = "female";
+const COED = "coed";
```
Brainicism: Might be cleaner to save these under an object like `GENDER.MALE`, `GENDER.FEMALE`, etc

game_session.js
```diff
@@ -63,4 +66,48 @@ module.exports = class GameSession {
     getDefaultBeginningCutoffYear() {
         return BEGINNING_SEARCH_YEAR;
     }
+
+    resetGender() {
```
Brainicism: Do we need to have a reset function for each individual parameter? Why not just reset everything at once.

Brainicism: If user enters an invalid input, just ignore it and keep the old values

game_session.js
```diff
@@ -63,4 +66,48 @@ module.exports = class GameSession {
...
+        this._gameOptions.gender = [ FEMALE ];
+    }
+
+    setGender(genderArr) {
+        // Return true when gender is successfully updated, false otherwise
+        if (genderArr.length === 0) {
+            this.resetGender();
+        }
+        else {
+            genderArr.map((gender, i) => { genderArr[i] = genderArr[i].toLowerCase(); })
```
Brainicism: `map()` doesn't mutate, and you're not doing anything with the return value

game_session.js
```diff
@@ -63,4 +66,48 @@ module.exports = class GameSession {
...
+            this._gameOptions.gender = [];
+            for (let i = 0; i < genderArr.length; i++) {
+                if (!this._gameOptions.gender.includes(genderArr[i]) &&
```
Brainicism: Can replace this with `!this._gameOptions.gender.includes(genderArr[i]) && Object.keys(GENDER).includes(genderArr[i])`

Brainicism: this probably works too
```
this._gameOptions = genderArr.filter((gender, pos) => { 
     gender = gender.toLowerCase();
     let firstOccurence = genderArr.indexOf(gender) == pos;
     return firstOccurence && Object.keys(GENDER).includes(gender);
})
```

game_session.js
```diff
@@ -63,4 +66,48 @@ module.exports = class GameSession {
...
+            }
+        }
+        return true;
+    }
+
+    getSQLGender() {
+        let genderStr = "";
+        this._gameOptions.gender.map((gender, i) => {
+            genderStr += `"${gender}"`;
+            if (this._gameOptions.gender.length !== i + 1) {
```
Brainicism: `i !== this._gameOptions.gender.length - 1` is more clear

game_session.js
taahamahdi: At this point I don't see the value in having a function to reset all options at once, but it might be valuable in the future (something like `!reset` maybe?). I've created reset functions to prevent code duplication (and for future-proofing if we have to do more advanced tasks in order to reset an option). Then, if we need to create a function to reset everything at once in the future, we can just call all of these helper functions and keep everything modular.

That being said, I can definitely implement keeping old values in the case of invalid input. Are you fine with me making a temporary gender array to store the old values so we have something to reset it to?

game_session.js
taahamahdi: Are you sure? I'm mutating the array directly. Before this, entering something like `MALE` wouldn't be identified.

game_session.js
Brainicism: Yeah I intended it for `!reset`, I just don't see any reason to auto-reset a parameter if an invalid input is entered.   

It'd probably be better to create a temporary array to store the new values, then assigning it to the actual gender array afterwards (assuming no invalid input)

game_session.js
Brainicism: Didn't see you were modifying the array contents in the callback, looks fine

game_session.js
```diff
@@ -63,4 +64,49 @@ module.exports = class GameSession {
...
+                    // Prevent duplicates and allow valid genders only
+                    tempArr.push(genderArr[i]);
+                }
+            }
+            if (tempArr.length === 0) {
+                // User gave invalid inputs only
+                return false;
+            }
+            else {
+                this._gameOptions.gender = tempArr;
```
Brainicism: Don't need to wrap this in an `else`

kmq.js
```diff
@@ -75,7 +75,30 @@ client.on("message", (message) => {
                 message.channel.send(`The new cutoff year is \`${gameSession.getBeginningCutoffYear()}\`.`);
             }
         }
+        else if (command.action === "gender") {
+            if (gameSession.setGender(command.components)) {
```
Brainicism: You can just have `setGender` return the gender array instead of exposing `getGenderArray`

taahamahdi: That'll make handling invalid input a bit harder, as I won't be able to differentiate between `!gender female` and garbage input.

Brainicism: You can return an empty array if input is garbage, then this `if` statement won't execute

PR #49, opened 2020-02-21
!end should not send a message if a game has not started [#48]

taahamahdi:


commands/end.js
```diff
@@ -9,10 +9,10 @@ module.exports = {
             message.channel.send(gameSession.scoreboard.getWinnerMessage());
             sendScoreboard(message, gameSession);
         }
-        else {
+        else if (gameSession.gameInSession()) {
```
Brainicism: Just add a check for `gameSession.gameInSession()` at the beginning of the function and return instead

PR #50, opened 2020-02-21
Updated SQL query to match MySQL syntax [#47]

taahamahdi:


models/game_session.js
```diff
@@ -96,7 +96,7 @@ module.exports = class GameSession {
         this._gameOptions.gender.map((gender, i) => {
             genderStr += `${gender}`;
             if (i !== this._gameOptions.gender.length - 1) {
-                genderStr += " OR ";
+                genderStr += ",";
```
Brainicism: This entire function can just be replaced with a .join(",")

PR #71, opened 2020-03-05
Fixes #67

### LINKED ISSUE #67: Sanitize accents from song names
A user reported requiring to enter `Décalcomanie` instead of `Decalcomanie` to get the song right.

On another note, consider keeping list of frequently used shortened versions of words, then replacing the user's guess with them (i.e. `u -> you`)

DannyKong12:
Normalizes string before removing non-ascii

.replace(/[^\x00-\x7F|]/g, "")

will remove é so Décalcomanie => dcalcomanie when decalcomanie is expected.

### DIFF
--- a/helpers/utils.js
+++ b/helpers/utils.js
@@ -64,7 +64,12 @@ module.exports = {
         return `${user.username}#${user.discriminator}`
     },
     cleanSongName: (name) => {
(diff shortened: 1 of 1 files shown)


DannyKong12:
String.prototype.normalize("NFD") changes "é" to "e´" and then the replace removes the "´" character

Issue #76, opened 2020-03-05
!stop when all users leave voice channel


taahamahdi:
The bot should execute the functions that !stop does when it is the only member left in VC.

PR #79, opened 2020-03-07
End round when all users leave VC [#76]

taahamahdi:
Can't send the song message because there is no way for us to get the text channel

### DIFF
--- a/kmq.js
+++ b/kmq.js
@@ -36,6 +36,24 @@ client.on("message", (message) => {
     }
 });
 
+client.on("voiceStateUpdate", (oldState, newState) => {
+    let oldUserChannel = oldState.voiceChannel;
+    let newUserChannel = newState.voiceChannel;
+    if (newUserChannel === undefined) {
+        // User left voice channel, check if bot is only one left
+        if (oldUserChannel.members.size === 1) {
+            let guildID = oldUserChannel.guild.id;
+            let voiceConnection = client.voiceConnections.get(guildID);
+            if (voiceConnection) {
+                voiceConnection.disconnect();
+                let gameSession = gameSessions[guildID];
+                gameSession.endRound();
+                return;
+            }
+        }
+    }
+});
+
 const parseMessage = (message, botPrefix) => {
     if (message.charAt(0) !== botPrefix) return null;
     let components = message.split(" ");


kmq.js
```diff
@@ -36,6 +36,25 @@ client.on("message", (message) => {
     }
 });
 
+client.on("voiceStateUpdate", (oldMember, newMember) => {
```
Brainicism: The parameters to the callback are `oldVoiceState` and `newVoiceState`, not specific to member

kmq.js
```diff
@@ -36,6 +36,25 @@ client.on("message", (message) => {
...
+    let oldUserChannel = oldMember.voiceChannel;
+    let newUserChannel = newMember.voiceChannel;
+    if (newUserChannel === undefined) {
+        // User left voice channel, check if bot is only one left
+        if (oldUserChannel.members.size === 1) {
+            let guildID = oldUserChannel.guild.id;
+            let voiceConnection = client.voiceConnections.get(guildID);
+            if (voiceConnection) {
+                sendSongMessage(message, gameSession, true);
```
Brainicism: Where is `message` coming from

kmq.js
```diff
@@ -36,6 +36,25 @@ client.on("message", (message) => {
     }
 });
 
+client.on("voiceStateUpdate", (oldMember, newMember) => {
+    let oldUserChannel = oldMember.voiceChannel;
+    let newUserChannel = newMember.voiceChannel;
```
Brainicism: Documentation says that this should be `.channel`, are you sure this works?

kmq.js
```diff
@@ -36,6 +36,25 @@ client.on("message", (message) => {
...
+    if (newUserChannel === undefined) {
```
Brainicism: Why is the voicestate undefined? I'm struggling to understand how this works, especially given that the bot is still in the voice channel.

kmq.js
```diff
@@ -36,6 +36,25 @@ client.on("message", (message) => {
...
+        // User left voice channel, check if bot is only one left
+        if (oldUserChannel.members.size === 1) {
+            let guildID = oldUserChannel.guild.id;
+            let voiceConnection = client.voiceConnections.get(guildID);
```
Brainicism: Voicestate should contain the corresponding connection already

kmq.js
taahamahdi: Yup! I tried with `.channel` and it doesn't work.

Issue #87, opened 2020-04-11
Implement volume control


taahamahdi:
Currently we hard-code the stream volume to 0.1. We should allow the user to modify this value from 0 to 0.1, and scale that as a percentage.

PR #89, opened 2020-04-11
Implemented volume control [#87]

taahamahdi:
- Added missing help string for prefix


commands/volume.js
```diff
@@ -0,0 +1,25 @@
+DEFAULT_VOLUME = 0.1;
```
Brainicism, line 1: missing `const`?

Brainicism: Shouldn't default be 50, if the random is bounded by [1,100]?

PR #99, opened 2020-04-17
Implemented caching songs when played [#9]

taahamahdi:


helpers/utils.js
```diff
@@ -1,7 +1,10 @@
 const RED = 0xE74C3C;
+const CACHE_DIR = "./.cache";
```
Brainicism, line 2: Should specify the cache directory in `config.json.` Also name the variable something more descriptive, something like song_cache_dir

helpers/utils.js
```diff
@@ -91,11 +94,46 @@ module.exports = {
...
+    if (!fs.existsSync(CACHE_DIR)) {
+        fs.mkdirSync(dir)
+    }
+
+    const ytdlOptions = {
+        filter: "audioonly",
+        quality: "highest"
+    };
+
+    const songLocation = `${CACHE_DIR}/${gameSession.getLink()}.mp3`;
```
Brainicism, line 116: cachedSongLocation

Brainicism: Another issue not related to this PR: 
`GameSession::getLink()` is misleading, it returns a youtube video ID, not the full URL. Could you refactor this?

helpers/utils.js
```diff
@@ -91,11 +94,46 @@ module.exports = {
...
+    if (!fs.existsSync(songLocation)) {
+        gameSession.isSongCached = false;
+        const tempLocation = `${songLocation}.part`;
+        if (!fs.existsSync(tempLocation)) {
+            let cacheStream = fs.createWriteStream(tempLocation);
+            ytdl(gameSession.getLink(), ytdlOptions)
+                .pipe(cacheStream);
+            cacheStream.on('finish', () => {
+                fs.renameSync(tempLocation, songLocation)
```
Brainicism, line 125: Can you create an issue to verify that the song was downloaded in full?

helpers/utils.js
```diff
@@ -91,11 +94,46 @@ module.exports = {
...
+    }
+
+    const ytdlOptions = {
+        filter: "audioonly",
+        quality: "highest"
+    };
+
+    const songLocation = `${CACHE_DIR}/${gameSession.getLink()}.mp3`;
+    if (!fs.existsSync(songLocation)) {
+        gameSession.isSongCached = false;
```
Brainicism, line 118: You can refactor this by doing `gameSession.isSongCached = !fs.existsSync(songLocation)` outside the `if`, then using that condition to check

helpers/utils.js
```diff
@@ -91,11 +94,46 @@ module.exports = {
...
+        const tempLocation = `${songLocation}.part`;
+        if (!fs.existsSync(tempLocation)) {
```
Brainicism, line 120: What happens if we have a partially completed download that was interrupted?

helpers/utils.js
taahamahdi: Sure and sure!

helpers/utils.js
taahamahdi: This needs to be fixed with that new issue you told me to create. It'll just stream the song instead, but the `.part` file will exist in an incomplete state.

PR #100, opened 2020-04-18
Separate volumes for streaming from YT and cached playback

taahamahdi:


helpers/utils.js
```diff
@@ -109,9 +114,8 @@ const playSong = (gameSession, guildPreference, db, message) => {
     };
 
     const songLocation = `${CACHE_DIR}/${gameSession.getLink()}.mp3`;
-    let isSongCached = true;
     if (!fs.existsSync(songLocation)) {
-        isSongCached = false;
+        gameSession.isSongCached = false;
```
Brainicism, line 118: Why does `isSongCached` need to be part of `GameSession`? Does it need to exist beyond the scope of this function?

taahamahdi: Yup, in volume.js. It's necessary for changing volume while playing back.

Issue #94, opened 2020-04-13
End round when an admin disconnects the bot


taahamahdi:
When a user tries to start the bot again, they get the warning the game is already in session. A fix for this will probably be inside `voiceStateUpdate`.

PR #106, opened 2020-04-20
End round when the bot is disconnected by admin [#94]

taahamahdi:


kmq.js
```diff
@@ -60,6 +60,12 @@ client.on("voiceStateUpdate", (oldState, newState) => {
                 return;
             }
         }
+        // Bot was disconnected by another user
+        if (!oldUserChannel.members.find((user) => user === client)) {
```
Brainicism, line 64: Can replace this with `!oldUserChannel.members.has(client)`

kmq.js
```diff
@@ -60,6 +60,12 @@ client.on("voiceStateUpdate", (oldState, newState) => {
...
+            let gameSession = gameSessions[guildID];
```
Brainicism, line 65: Can move this outside the conditional and re-use it in both cases

PR #111, opened 2020-05-08
Added null check to gameSession in VCUpdate

taahamahdi:


kmq.js
```diff
@@ -56,14 +56,17 @@ client.on("voiceStateUpdate", (oldState, newState) => {
             let voiceConnection = client.voice.connections.get(guildID);
             if (voiceConnection) {
                 voiceConnection.disconnect();
-                gameSession.endRound();
+                if (gameSession) {
```
Brainicism, line 59: Why not check for `!gameSession` on line 54 and do early return instead

taahamahdi: Then `.disconnect()` won't be called, right?

taahamahdi: Whoops, I don't think it should be possible.

Issue #46, opened 2020-02-20
Add command to show current game session options


Brainicism:

taahamahdi:
I think this command should also be called when the user pings the bot. That way, if the user forgot their prefix, they can find out what it is.

taahamahdi:
Leave this one open a bit longer, I'm adding a footer to show the bot prefix

PR #125, opened 2020-05-10
Options give prefix, options on bot ping [#46]

taahamahdi:


helpers/utils.js
```diff
@@ -47,18 +48,22 @@ const sendSongMessage = (message, gameSession, isForfeit) => {
...
-    message.channel.send({
-        embed: {
-            color: EMBED_INFO_COLOR,
-            author: {
-                name: message.author.username,
-                icon_url: message.author.avatarURL()
-            },
-            title: `**${title}**`,
-            description: description
+const sendInfoMessage = (message, title, description, footerText, footerImageWithPath, footerImage) => {
```
Brainicism, line 51: What happens if footer isn't specified?

Brainicism: Why does need footer image with path and footer image

taahamahdi: If they aren't specified nothing happens. They're optional and the embed behaves as it should when they aren't given, without any errors.

taahamahdi: The reason I need to differentiate with the path and the regular filename is because the embed needs the path to the file when adding an attachment, but when being used in the embed, it needs only the filename. I can actually get away with just passing the path by using something along the lines of:
```js
var path = require('path');
var file = '/home/user/dir/file.txt';

var filename = path.parse(file).base;
```

helpers/utils.js
```diff
@@ -48,18 +51,25 @@ const sendSongMessage = (message, gameSession, isForfeit) => {
...
+    let footerImage = footerImageWithPath ? path.basename(footerImageWithPath) : null;
+    let embed = new Discord.MessageEmbed({
+        color: EMBED_INFO_COLOR,
+        author: {
+            name: message.author.username,
+            icon_url: message.author.avatarURL()
+        },
+        title: `**${title}**`,
+        description: description,
+        footer: {
```
Brainicism, line 64: You can attach the `footer` afterwards in the if statement afterwards

taahamahdi: Awesome, just did it!

helpers/utils.js
```diff
@@ -48,18 +51,24 @@ const sendSongMessage = (message, gameSession, isForfeit) => {
...
-            author: {
-                name: message.author.username,
-                icon_url: message.author.avatarURL()
-            },
-            title: `**${title}**`,
-            description: description
-        }
-    });
+const sendInfoMessage = (message, title, description, footerText, footerImageWithPath) => {
+    let footerImage = footerImageWithPath ? path.basename(footerImageWithPath) : null;
```
Brainicism, line 55: Can move this into the if statement

taahamahdi: Will do

PR #138, opened 2020-05-14
Added functionality to post server count to top.gg [#130]

### LINKED ISSUE #130: Post server count to top.gg

taahamahdi:
Closes #130

### DIFF
--- a/kmq.js
+++ b/kmq.js
@@ -1,7 +1,8 @@
 const Discord = require("discord.js");
 const mysql = require("promise-mysql");
-
 const fs = require("fs");
+const DBL = require("dblapi.js");
+
 const client = new Discord.Client();
 const logger = require('./logger')("kmq");
 const config = require("./config.json");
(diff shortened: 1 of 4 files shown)


kmq.js
```diff
@@ -10,13 +11,15 @@ const guessSong = require("./helpers/guess_song");
...
 const { clearPartiallyCachedSongs } = require("./helpers/utils");
+
 let db;
 let commands = {};
 let gameSessions = {};
 let guildPreferences = {};
 
 client.on("ready", () => {
     logger.info(`Logged in as ${client.user.tag}!`);
+    postTopGGCount();
```
Brainicism, line 22: This needs to happen outside of `.on("ready")` because the top.gg API waits for the same event

kmq.js
```diff
@@ -99,6 +103,20 @@ const parseMessage = (message, botPrefix) => {
     }
 }
 
+const hookTopGGCount = () => {
+    if (!config.topGGToken) {
+        logger.info("No top.gg token passed (check your config.json)! Ignoring posting top.gg server count.");
+        return;
+    }
+    const dbl = new DBL(config.topGGToken, client);
```
Brainicism, line 111: This will go out of scope and get garbage collected
https://stackoverflow.com/a/38449557

taahamahdi: I removed the function. I also refactored this part a bit so that it shouldn't be garbage collected anymore.

PR #160, opened 2020-05-21
Only stream cached songs

Brainicism:


src/kmq.ts
```diff
@@ -42,6 +42,7 @@ client.on("ready", () => {
 
 client.on("message", (message: Discord.Message) => {
     if (message.author.equals(client.user) || message.author.bot) return;
+    message.
```
taahamahdi, line 45: Don't think you meant to leave this here

src/scripts/download-new-songs.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+}
+
+const downloadSong = (id: string) => {
+    let cachedSongLocation = path.join(config.songCacheDir, `${id}.mp3`);
+    const tempLocation = `${cachedSongLocation}.part`;
+    let cacheStream = fs.createWriteStream(tempLocation);
+    const ytdlOptions = {
+        filter: "audioonly" as const,
+        quality: "highest"
```
taahamahdi, line 40: Can this be as const as well?

Brainicism: The `const` cast is only required because `ytdlOptions.filter` requires type `Filter` which is defined as `'audioandvideo' | 'video' | 'videoonly' | 'audio' | 'audioonly' | ((format: videoFormat) => boolean);`, it will not compile otherwise

taahamahdi: Understood, thanks!

PR #185, opened 2020-05-30
Ignore non-bot voiceStateUpdates when checking if bot was kicked

taahamahdi:


src/kmq.ts
```diff
@@ -74,7 +74,7 @@ client.on("message", (message: Discord.Message) => {
 
 client.on("voiceStateUpdate", async (oldState, newState) => {
     let oldUserChannel = oldState.voiceChannel;
-    if (!oldUserChannel) {
+    if (!oldUserChannel || oldState.user !== client.user) {
```
Brainicism, line 77: We still need to listen to user voice state changes for the the `User left voice channel, check if bot is only one left` case

taahamahdi: Thanks, can you take a look again? I made it how you originally suggested.

Issue #188, opened 2020-06-17
Delay between ,skip call and actually playing next song


taahamahdi:
Bot was running for a while, so this might be a hard one to reproduce.

taahamahdi:
A possible related issue: sometimes the correct guess noise doesn't play, and instead a small snippet of the next song plays for less than a second.

PR #196, opened 2020-07-25
Fixed bug where songs would continue playing after game ended

Brainicism:


taahamahdi:
Does this fix #188 as well?

Brainicism:
I don't think so

taahamahdi:
Alright

Issue #131, opened 2020-05-11
Add usage data to guild table


Brainicism:
Date of first use, last active, etc.

PR #206, opened 2020-07-28
Update activity timestamp on `,play`, `,stop`, `,skip`, and guess [#198]

### LINKED ISSUE #198: Different metric for game session inactivity
Game sessions are marked inactive based on the last time `startRound` and `endRound` are called internally. This can happen even if users afk and let the songs continuously play

taahamahdi:
- Don't need to update on `,end` since GameSession gets destroyed anyway
- Now we only process guesses when game is in session, not processing when game is stopped

Closes #198

### DIFF
--- a/src/commands/play.ts
+++ b/src/commands/play.ts
@@ -18,6 +18,7 @@ class PlayCommand implements BaseCommand {
             let channel = message.channel as TextChannel;
             if (!gameSessions[message.guild.id]) {
                 gameSessions[message.guild.id] = new GameSession(channel);
+                // GameSession constructor sets lastActive to current time during constructor
                 logger.info(`${getDebugContext(message)} | Game session created`);
             }
             await sendInfoMessage(message, `Game starting in #${channel.name}`, "Listen to the song and type your guess!");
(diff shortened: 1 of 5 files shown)


src/commands/play.ts
```diff
@@ -18,6 +18,7 @@ class PlayCommand implements BaseCommand {
             let channel = message.channel as TextChannel;
             if (!gameSessions[message.guild.id]) {
                 gameSessions[message.guild.id] = new GameSession(channel);
+                // GameSession constructor sets lastActive to current time during constructor
```
Brainicism, line 21: i don't think this comment is necessary

src/commands/skip.ts
```diff
@@ -31,6 +31,7 @@ class SkipCommand implements BaseCommand {
             await gameSession.endRound();
             startGame(gameSession, guildPreference, db, message, client);
             logger.info(`${getDebugContext(message)} | Skip majority achieved.`);
+            gameSession.lastActive = Date.now();
```
Brainicism, line 34: Do we only want to consider activity when skip majority is achieved? Or when a skip vote is received too

taahamahdi: I'm fine either way. What do you prefer?

Brainicism: think it should be on both

taahamahdi: I'll update accordingly

Brainicism:
Can you also make these part of a method on `GameSession.` Might re-use this to update server activity statistics for [#131]

PR #210, opened 2020-07-30
Remove delay between ,skip and next song start [#188]

### LINKED ISSUE #188: Delay between ,skip call and actually playing next song

taahamahdi:
Closes #188


src/commands/skip.ts
```diff
@@ -26,7 +26,7 @@ class SkipCommand implements BaseCommand {
         }
         if (isSkipMajority(message, gameSession)) {
             gameSession.skipAchieved = true;
-            await sendSkipMessage(message, gameSession);
+            sendSkipMessage(message, gameSession);
```
Brainicism, line 29: Still want consistent ordering of `sendSkipMessage` followed by `sendSongMessage.` We can change the implementation of `sendSkipMessage` to `await` the first promise to retrieve the new message, then call `.delete` on it without `await`

taahamahdi: Good thinking!

PR #229, opened 2020-08-08
Workaround for bug where delay between songs continuously increased

Brainicism:
Increasing delay between `voiceConnection.play()` calls resets every time we re-join the voice connection. GameSession automatically leaves and rejoins the channel every 15 seconds as a workaround.

### DIFF
--- a/src/kmq.ts
+++ b/src/kmq.ts
@@ -86,13 +86,6 @@ client.on("voiceStateUpdate", async (oldState, newState) => {
                 return;
             }
         }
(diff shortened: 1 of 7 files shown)


src/kmq.ts
```diff
@@ -86,13 +86,6 @@ client.on("voiceStateUpdate", async (oldState, newState) => {
                 return;
             }
         }
-        // Bot was disconnected voice channel (either via a kick by an admin or the situation handled above)
```
taahamahdi, line 89: So we don't need this logic anymore?

PR #250, opened 2020-08-15
Migrate to Eris discord framework

Brainicism:


src/helpers/bot_stats_poster.ts
```diff
@@ -1,4 +1,4 @@
-import { Client } from "discord.js";
```
taahamahdi, line 1: Couldn't you make changing this file less work by just importing client from eris?

Brainicism: Imports in node are cached, so every file importing will receive the same reference. We do `import * as Eris` in a bunch of a files, so doing it here wouldn't affect anything

src/models/guild_preference.ts
```diff
@@ -177,7 +177,7 @@ export default class GuildPreference {
     async updateGuildPreferences(db: Knex) {
         await db("guild_preferences")
             .where({ guild_id: this.guildID })
-            .update({ guild_preference: JSON.stringify(this) });
+            .update({ guild_preference: JSON.stringify(this)});
```
taahamahdi, line 180: Hey I get it you don't play by the rules Mr bigshot

PR #255, opened 2020-08-16
Only select songs from those have already been downloaded

Brainicism:


src/helpers/game_utils.ts
```diff
@@ -185,26 +190,8 @@ async function selectRandomSong(guildPreference: GuildPreference): Promise<Queri
...
-        }
-        const songLocation = `${SONG_CACHE_DIR}/${random.youtubeLink}.mp3`;
-        if (!fs.existsSync(songLocation)) {
-            logger.error(`Song not cached: ${songLocation}`);
-            attempts++;
-            continue;
-        }
-        return random;
-    }
+    const random = queriedSongList[Math.floor(Math.random() * queriedSongList.length)];
```
taahamahdi, line 193: Could just return instead of assigning a variable, I don't think it's being used anywhere else

src/scripts/download-new-songs.ts
```diff
@@ -84,35 +85,35 @@ const downloadSong = (id: string) => {
...
-        query = `SELECT nome as name, name as artist, vlink as youtubeLink FROM kpop_videos.app_kpop INNER JOIN kpop_videos.app_kpop_group ON kpop_videos.app_kpop.id_artist = kpop_videos.app_kpop_group.id
-        WHERE dead = "n" AND vtype = "main" ORDER BY kpop_videos.app_kpop.views DESC LIMIT ${limit};`;
+        songs = songs.slice(0, limit);
     }
-    const songs: Array<QueriedSong> = await db.query(query);
     let downloadCount = 0;
     logger.info("Total songs in database: " + songs.length);
-    const songsNotDownloaded = songs.filter(x => !fs.existsSync(path.join(config.songCacheDir, `${x.youtubeLink}.mp3`)));
-    logger.info("Total songs to be downloaded: " + songsNotDownloaded.length);
+    const songsToDownloaded = songs.filter(x => !fs.existsSync(path.join(config.songCacheDir, `${x.youtubeLink}.mp3`)));
```
taahamahdi, line 107: songsToDownload is what I think you meant maybe?

PR #259, opened 2020-08-16
Add number of recent games KMQ fact

Brainicism:


src/fact_generator.ts
```diff
@@ -388,3 +388,13 @@ async function genderGamePreferences(): Promise<string[]> {
     const femaleProportion = (100 * (femaleCount / preferenceCount)).toFixed(2);
     return [`KMQ Fact: ${femaleProportion}% of servers play with only girl group songs, while ${maleProportion}% play with boy groups only!`]
 }
+
+async function recentGames(): Promise<string[]> {
+    const oneWeekPriorDate = new Date();
+    oneWeekPriorDate.setDate(oneWeekPriorDate.getDate() - 7);
```
taahamahdi, line 394: Since you are setting the date, shouldn't this be mutable?

PR #263, opened 2020-08-17
Convert mp3 files to opus for less CPU usage (hopefully)

taahamahdi:


taahamahdi:
Breaks volume control entirely.

PR #297, opened 2020-09-09
Use dotenv instead of config json

Brainicism:
- Use dotenv instead of config json
- Modify tsconfig to not include src/ in build/ to make directory structure level with both tsc and ts-node


taahamahdi (approved):
###

PR #309, opened 2020-10-12
Add server join message in a debug channel

taahamahdi:
Not user-facing. We can use this debug channel for other information in the future.

### DIFF
--- a/src/events/client/guildCreate.ts
+++ b/src/events/client/guildCreate.ts
@@ -1,0 +1,34 @@
+import Eris from "eris";
+import _logger from "../../logger";
+import { state } from "../../kmq";
+import { sendMessage, EMBED_INFO_COLOR } from "../../helpers/discord_utils";
+const logger = _logger("guildCreate");
+
+export default async function guildCreateHandler(guild: Eris.Guild) {
+    const client = state.client;
+    logger.info(`New server joined: ${guild.id} with ${guild.memberCount} users`);
+    if (!process.env.DEBUG_SERVER_ID || !process.env.DEBUG_TEXT_CHANNEL_ID) return;
+    const kmqDebugChannel: Eris.TextChannel =
+        <Eris.TextChannel>client.guilds.get(process.env.DEBUG_SERVER_ID)
+            .channels.get(process.env.DEBUG_TEXT_CHANNEL_ID);
+    const joinDate: Date = new Date(guild.joinedAt);
+    await sendMessage({ channel: kmqDebugChannel }, {
+        embed: {
+            color: EMBED_INFO_COLOR,
+            author: {
+                name: guild.name,
+                icon_url: guild.iconURL
+            },
+            title: `New server joined!`,
+            fields: [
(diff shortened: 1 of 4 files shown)


src/events/client/guildCreate.ts
```diff
@@ -0,0 +1,34 @@
...
+import _logger from "../../logger";
+import { state } from "../../kmq";
+import { sendMessage, EMBED_INFO_COLOR } from "../../helpers/discord_utils";
+const logger = _logger("guildCreate");
+
+export default async function guildCreateHandler(guild: Eris.Guild) {
+    const client = state.client;
+    logger.info(`New server joined: ${guild.id} with ${guild.memberCount} users`);
+    const kmqDebugChannel: Eris.TextChannel = <Eris.TextChannel>client.guilds.find((guild: Eris.Guild) => {
+        if (guild.name === "KMQ") {
```
Brainicism, line 11: Can get rid of the `if` and just do a single line return on the condition. 
Compare guild.id instead of name

src/events/client/guildCreate.ts
```diff
@@ -0,0 +1,34 @@
...
+            return true;
+        }
+    }).channels.find((channel: Eris.AnyGuildChannel) => {
+        if (channel.name === "kmq-debug") {
```
Brainicism, line 15: Can get rid of the `if` and just do a single line return on the condition.
Compare channel.id instead of name

src/events/client/guildCreate.ts
```diff
@@ -0,0 +1,34 @@
...
+    const joinDate: Date = new Date(guild.joinedAt);
+    await sendMessage({ channel: kmqDebugChannel }, {
+        embed: {
+            color: EMBED_INFO_COLOR,
+            author: {
+                name: guild.name,
+                icon_url: guild.iconURL
+            },
+            title: `New server joined!`,
+            description: `**Member Count**: ${guild.memberCount}\n**Region**: ${guild.region}\n**Language**: ${guild.preferredLocale}\n**Nitro Boosts**: ${guild.premiumSubscriptionCount}`,
```
Brainicism, line 28: Use embed fields instead of multi-line description

src/events/client/guildCreate.ts
Brainicism: Use `.get` instead of `.find`

src/events/client/guildCreate.ts
Brainicism: Use `.get` instead of `.find`

src/events/client/guildCreate.ts
Brainicism: Add these to config instead of hardcode

src/events/client/guildCreate.ts
Brainicism: Add these to config instead of hardcode

Brainicism (changes requested):

Brainicism (approved):

PR #334, opened 2020-11-06
Disallow users from setting limit when total song count is 0

### LINKED ISSUE #333: Limit can become 0
image

Repro:
1. Use filters to end up with 0 available songs, e.g `,groups rm`, `,cutoff 2020`
2. Set any limit, e.g `,limit 500`

Brainicism:
Closes #333


src/commands/game_options/limit.ts
```diff
@@ -12,6 +12,10 @@ export default class LimitCommand implements BaseCommand {
         const guildPreference = await getGuildPreference(message.guildID);
         guildPreference.setLimit(parseInt(parsedMessage.components[0], 10));
         const songCount = await getSongCount(guildPreference);
+        if (songCount === 0){
```
taahamahdi, line 15: Put space after the )

taahamahdi (changes requested):

taahamahdi (approved):

PR #377, opened 2020-11-16
updateGuildPreferences call informs whether available songs may change [#376]

taahamahdi:
is this better or worse than using true/false :thinking:

### DIFF
--- a/src/models/guild_preference.ts
+++ b/src/models/guild_preference.ts
@@ -66,18 +66,20 @@ export default class GuildPreference {
             }
         }
         if (gameOptionModified) {
-            this.updateGuildPreferences(dbContext.kmq, false);
+            this.updateGuildPreferences(dbContext.kmq);
         }
     }
 
     setLimit(limit: number) {
         this.gameOptions.limit = limit;
-        this.updateGuildPreferences(dbContext.kmq, true);
+        this.updateGuildPreferences(dbContext.kmq);
+        this.updateGameSession(true);
     }
 
     resetLimit() {
         this.gameOptions.limit = DEFAULT_LIMIT;
-        this.updateGuildPreferences(dbContext.kmq, true);
+        this.updateGuildPreferences(dbContext.kmq);
+        this.updateGameSession(true);
     }
 
     getLimit(): number {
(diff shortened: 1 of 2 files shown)


src/models/guild_preference.ts
```diff
@@ -38,6 +38,11 @@ interface GameOptions {
     guessTimeout: number;
 }
 
+const enum AvailableSongs {
```
Brainicism, line 41: Use boolean

src/models/guild_preference.ts
```diff
@@ -223,11 +231,13 @@ export default class GuildPreference {
         return this.gameOptions.shuffleType === ShuffleType.UNIQUE;
     }
 
-    async updateGuildPreferences(_db: Knex) {
+    async updateGuildPreferences(_db: Knex, songsState: AvailableSongs) {
```
Brainicism, line 234: songsState => `songListModified`

src/models/guild_preference.ts
```diff
@@ -223,11 +231,13 @@ export default class GuildPreference {
...
         await _db("guild_preferences")
             .where({ guild_id: this.guildID })
             .update({ guild_preference: JSON.stringify(this) });
         const gameSession = state.gameSessions[this.guildID];
-        if (gameSession) gameSession.resetLastPlayedSongsQueue();
+        if (gameSession && songsState === AvailableSongs.Modifies) {
+            gameSession.resetLastPlayedSongsQueue();
```
Brainicism, line 240: Don't we still have this logic in gamesession

Brainicism (changes requested):

src/models/guild_preference.ts
taahamahdi: That call to reset is when the user plays all of their unique songs, so it's still needed

src/models/guild_preference.ts
```diff
@@ -67,17 +67,20 @@ export default class GuildPreference {
         }
         if (gameOptionModified) {
             this.updateGuildPreferences(dbContext.kmq);
+            this.updateGameSession(false);
```
Brainicism, line 70: Game session doesn't exist at this point, not needed here

Brainicism (changes requested):

src/models/guild_preference.ts
taahamahdi: Thanks!

Brainicism (approved):

PR #379, opened 2020-11-16
Use default options when game options with no args [#370]

### LINKED ISSUE #370: Using game options without parameters should reset them to default values

taahamahdi:
Closes #370.

### DIFF
--- a/src/commands/game_options/cutoff.ts
+++ b/src/commands/game_options/cutoff.ts
@@ -5,16 +5,24 @@ import _logger from "../../logger";
 import { GameOption } from "../../types";
 
 const logger = _logger("cutoff");
-export const BEGINNING_SEARCH_YEAR = 2008;
+export const DEFAULT_BEGINNING_SEARCH_YEAR = 2008;
+export const DEFAULT_ENDING_SEARCH_YEAR = (new Date()).getFullYear();
 
 export default class CutoffCommand implements BaseCommand {
     async call({ message, parsedMessage }: CommandArgs) {
         const guildPreference = await getGuildPreference(message.guildID);
+        if (parsedMessage.components.length === 0) {
+            guildPreference.setBeginningCutoffYear(DEFAULT_BEGINNING_SEARCH_YEAR);
+            guildPreference.setEndCutoffYear(DEFAULT_ENDING_SEARCH_YEAR);
+            await sendOptionsMessage(message, guildPreference, GameOption.CUTOFF);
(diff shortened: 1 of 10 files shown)


src/commands/game_options/cutoff.ts
```diff
@@ -11,8 +11,11 @@ export default class CutoffCommand implements BaseCommand {
     async call({ message, parsedMessage }: CommandArgs) {
         const guildPreference = await getGuildPreference(message.guildID);
         const yearRange = parsedMessage.components;
-        const startYear = yearRange[0];
-        if (yearRange.length === 1) {
+        const startYear = yearRange[0] || null;
```
Brainicism, line 14: What's the purpose of the `|| null` here?

src/models/guild_preference.ts
```diff
@@ -15,11 +15,11 @@ const logger = _logger("guild_preference");
 const DEFAULT_OPTIONS = {
     beginningYear: BEGINNING_SEARCH_YEAR,
```
Brainicism, line 16: Rename BEGINNING_SEARCH_YEAR to DEFAULT_BEGINNING_SEARCH_YEAR

src/models/guild_preference.ts
```diff
@@ -15,11 +15,11 @@ const logger = _logger("guild_preference");
...
     endYear: (new Date()).getFullYear(),
```
Brainicism, line 17: Make this a constant and name if DEFAULT_ENDING_SEARCH_YEAR

Brainicism (changes requested):

src/commands/game_options/cutoff.ts
taahamahdi: parsedMessage.components at 0 may be null when no arguments are passed

taahamahdi: Am I using it wrong?

Brainicism: You already have a `yearRange.length === 0` check below, so it would never be null where it matters. Alternatively, make a this `=== 0` check a separate if statement with an early return, then evaluate `startYear` after

Brainicism (changes requested):

src/commands/game_options/cutoff.ts
taahamahdi: My mistake, thanks for the explanation! I'll do the early return

Brainicism (approved):

PR #420, opened 2020-12-14
Update ytdl-core to fix hanging issue

Brainicism:


taahamahdi (approved):
Fixes that last case of needing to download and encode at once hanging.

PR #432, opened 2020-12-30
End game on skip and forceskip

taahamahdi:
Check for gameFinished inside each endRound call

- Replaces the check inside setTimeout
- No longer need the check inside ,forceskip and ,skip

### DIFF
--- a/src/models/game_session.ts
+++ b/src/models/game_session.ts
@@ -74,7 +74,7 @@ export default class GameSession {
 
     constructor(textChannel: Eris.TextChannel, voiceChannel: Eris.VoiceChannel, gameSessionCreator: Eris.User, gameType: GameType, eliminationLives?: number) {
         this.gameType = gameType;
-        this.scoreboard = this.gameType === GameType.ELIMINATION ? new EliminationScoreboard(eliminationLives) : new Scoreboard();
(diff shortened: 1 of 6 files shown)


src/models/game_session.ts
```diff
@@ -104,6 +104,13 @@ export default class GameSession {
             this.connection.removeAllListeners();
         }
         this.stopGuessTimeout();
+        if (this.gameType === GameType.ELIMINATION) {
+            const eliminationScoreboard = this.scoreboard as EliminationScoreboard;
```
Brainicism, line 108: Do we need this cast?

Brainicism: We can make the `gameFinished` check for classic mode in here as well

Brainicism (changes requested):

src/models/game_session.ts
taahamahdi: The cast was needed to call `eliminationScoreboard`'s `gameFinished` method instead of `Scoreboard`'s

taahamahdi: > We can make the `gameFinished` check for classic mode in here as well

To do this, we need to access `guildPreference` for `isGoalSet`, so we need to call `getGuildPreference`. However, to await this, I need to turn endRound into an asynchronous method. Is that okay?

Brainicism (approved):

PR #435, opened 2021-01-01
GAON weekly facts + fact cache generation rework

Brainicism:


src/fact_generator.ts
```diff
@@ -452,3 +476,38 @@ async function mostGaonAppearances(): Promise<string[]> {
         .limit(25);
     return result.map((x, idx) => `Fun Fact: '${x.artist_name}' has placed on the GAON digital weekly charts the ${getOrdinalNum(idx + 1)} most times with ${x.appearances} appearances!`);
 }
+
+async function historicalGaonWeekly(): Promise<Array<string>> {
+    const startYear = 2010;
+    const endYear = new Date().getFullYear() - 1;
```
taahamahdi, line 482: Can you explain why you're doing -1 here and +1 on line 489? I don't get it

src/helpers/utils.ts
```diff
@@ -94,3 +94,18 @@ export function arrayToString(arr: Array<string>): string {
...
+ */
+export function weekOfYear(dateObj?: Date): number {
+    const date = dateObj || new Date();
+    date.setHours(0, 0, 0, 0);
+    // Thursday in current week decides the year.
+    date.setDate(date.getDate() + 3 - ((date.getDay() + 6) % 7));
+    // January 4 is always in week 1.
+    const week1 = new Date(date.getFullYear(), 0, 4);
+    // Adjust to Thursday in week 1 and count number of weeks from date to week1.
+    return 1 + Math.round(((date.getTime() - week1.getTime()) / 86400000 - 3 + ((week1.getDay() + 6) % 7)) / 7);
```
taahamahdi, line 110: 😭

taahamahdi (approved):

src/fact_generator.ts
Brainicism: These quotes show songs for previous years, `endYear` evaluates to 2019 in 2020. Line 489 generates a range from 2010 to 2019, and the +1 creates the correct array size otherwise you'd have an off by one error

taahamahdi (approved):

PR #444, opened 2021-01-10
Add profile command

Brainicism:


taahamahdi (changes requested):
Looks great -- to add to this, users should be able to mention another user to retrieve their stats as well. What do you think?

data/end_game_messages.json
```diff
@@ -57,6 +57,10 @@
         {
             "title": "Did you know?",
             "message": "You can set a time limit for guessing each song by using the `timer` game option. See `,help timer` for more details."
+        },
+        {
+            "title": "Did you know?",
+            "message": "You can see how many games you've played, and how many song's you've guessed in total using `,profile`."
```
taahamahdi, line 63: you've played and how many songs

src/commands/game_commands/profile.ts
```diff
@@ -0,0 +1,62 @@
...
+import _logger from "../../logger";
+import { bold, friendlyFormattedDate } from "../../helpers/utils";
+
+const logger = _logger("profile");
+
+export default class ProfileCommand implements BaseCommand {
+    help = {
+        name: "profile",
+        description: "Shows your game stats.",
+        usage: "!profile",
```
taahamahdi, line 15: Should we include here that a username can be mentioned, or will it make it more confusing than it needs to be for the user?

taahamahdi (changes requested):

src/commands/game_commands/profile.ts
```diff
@@ -0,0 +1,69 @@
...
+    help = {
+        name: "profile",
+        description: "Shows your game stats.",
+        usage: "!profile { @mention }",
+        examples: [{
+            example: "`!profile`",
+            explanation: "View your own player profile.",
+        },
+        {
+            example: "`!shuffle @FortnitePlayer`",
```
taahamahdi, line 21: !profile

taahamahdi (changes requested):

taahamahdi (approved):

PR #549, opened 2021-02-14
Correctly check if bot is alone, also check if VC is empty

taahamahdi:


src/helpers/discord_utils.ts
```diff
@@ -453,16 +453,18 @@ export async function textPermissionsCheck(message: GuildTextableMessage): Promi
 
 /**
  * @param gameSession - The currently active GameSession
- * @param channel - The voice channel the bot is currently in
- * @returns whether the bot is alone 😔
+ * @param channel - The voice channel the bot could be in
+ * @returns whether the bot is alone 😔 ends the gameSession if it does
  */
-export async function checkBotIsAlone(gameSession: GameSession, channel: Eris.VoiceChannel) {
+export async function checkBotIsAlone(gameSession: GameSession, channel: Eris.VoiceChannel): Promise<boolean> {
```
Brainicism, line 459: This function isn't async anymore

src/structures/game_session.ts
```diff
@@ -357,6 +357,14 @@ export default class GameSession {
         this.prepareRound(randomSong.name, randomSong.artist, randomSong.youtubeLink);
         this.gameRound.setBaseExpReward(await this.calculateBaseExp(guildPreference));
 
+        if ((await checkBotIsAlone(this, this.voiceChannel))) {
+            return;
+        }
+        if (this.voiceChannel.voiceMembers.size === 0) {
```
Brainicism, line 363: Isn't this redundant

Brainicism (changes requested):

src/structures/game_session.ts
taahamahdi: This handles the case where before any rounds start the user leaves the voice channel. checkBotIsAlone only ends the session if the bot is the only one the voice channel (members == 1)

Brainicism (approved):

PR #566, opened 2021-03-01
Added python script to request latest view counts from YT

taahamahdi:
Updates KMQ database available_songs' views field

### DIFF
--- a/src/scripts/get-youtube-count.py
+++ b/src/scripts/get-youtube-count.py
@@ -1,0 +1,71 @@
+# Requires file named client_secret.json in same directory, containing
+# OAuth 2.0 Client ID with youtube.readonly scope in YouTube Data API v3
+# https://console.developers.google.com/
+#
+# Required modules:
+# python -m pip install
+# mysql
+# mysql-connector
+# python-dotenv
+# google-api-python-client
+# google-auth-oauthlib
+
+import os
+import json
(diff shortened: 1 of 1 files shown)


src/scripts/get-youtube-count.py
```diff
@@ -0,0 +1,71 @@
+# Requires file named client_secret.json in same directory, containing
```
Brainicism, line 1: Create own directory for this file, with requirements.txt

Brainicism: Add client_secret.json to gitignore

src/scripts/get-youtube-count.py
```diff
@@ -0,0 +1,71 @@
...
+db = mysql.connector.connect(
+    host=os.getenv("DB_HOST"),
+    user=os.getenv("DB_USER"),
+    password=os.getenv("DB_PASS"),
+    database=os.getenv("DB_KMQ_SETTINGS_TABLE_NAME")
+)
+
+cursor = db.cursor()
+
+cursor.execute("SELECT link FROM available_songs")
```
Brainicism, line 35: `available_songs` needs a `last_updated` field, and this query should only attempt to update songs that are `x` days old

Brainicism (changes requested):

PR #573, opened 2021-03-04
Add support for command hot-reloading

Brainicism:


src/helpers/management_utils.ts
```diff
@@ -229,24 +246,24 @@ export function getCommandFiles(): Promise<{ [commandName: string]: BaseCommand
...
+    if (commandName in state.commands && !reload) {
         logger.error(`Command \`${commandName}\` already exists. Possible conflict?`);
         process.exit(1);
     }
     state.commands[commandName] = command;
 }
 
 /** Registers commands */
-export async function registerCommands() {
+export async function registerCommands(reload = false) {
```
scarloxie, line 258: Reload seems like a bit of a misnomer here, just guessing here but seems like `registerCommands(false)` is only ever called on startup. Might as well assert that commands is empty when not reloading for clarity since it's effectively going to fail anyways.

Maybe im just boomer brain but it took me a second to understand whats going on here

scarloxie (approved):
approved with suggestions

taahamahdi (dismissed):

taahamahdi (approved):

PR #624, opened 2021-04-02
Add correct_guesses field to game-session-table

taahamahdi:


migrations/20210402030127_game-session-table.js
```diff
@@ -0,0 +1,11 @@
+exports.up = function(knex) {
```
Brainicism, line 1: Name of migration file could be more descriptive

Brainicism (approved):

PR #653, opened 2021-04-09
Add current players and game session count as fun facts

taahamahdi:


src/fact_generator.ts
```diff
@@ -415,6 +416,16 @@ async function recentUniquePlayers(): Promise<string[]> {
     return output;
 }
 
+async function currentUniquePlayers(): Promise<Array<string>> {
```
Brainicism, line 419: These facts are only refreshed once every 24 hours, the data would be very stale

Brainicism (changes requested):

src/fact_generator.ts
taahamahdi: Oh yeah, totally forgot about that. I'll close this PR

PR #718, opened 2021-05-01
Separate each group with a space in ,list

taahamahdi:
Improves readability substantially

### DIFF
--- a/src/structures/guild_preference.ts
+++ b/src/structures/guild_preference.ts
@@ -286,7 +286,7 @@ export default class GuildPreference {
     getDisplayedGroupNames(original = false): string {
         if (this.gameOptions.groups === null) return null;
         if (original) {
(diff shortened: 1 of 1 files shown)


Brainicism:
There's a lack of space intentionally to maximize the number of groups that can fit in the message, so it can be re-used with `,groups`

taahamahdi:
It is the only way for the user to see all the groups they have set if they go beyond what ,groups shows. Should we sacrifice readability for fitting in more groups when we already have the ability to send the groups list as a file if it's too long?

Brainicism:
There are cases where the text will fit in a `,groups` message when theres no space in the delimiter, but won't fit if there is a space. Unless we have two separate options to have space vs non-space, we can't satisfy both use cases at the same time

taahamahdi:
Understood, I'll close this PR

PR #754, opened 2021-05-16
Store each guild preference as separate entry in DB

taahamahdi:


src/scripts/json-presets-to-new-format.ts
```diff
@@ -0,0 +1,30 @@
+import { DatabaseContext, getNewConnection } from "../database_context";
+
+async function exportJsonPresetsToNewTable(db: DatabaseContext): Promise<void> {
+    const jsonPresets = await db.kmq("game_option_presets_json").select("*");
+    await Promise.all(jsonPresets.map(async (preset) => {
+        const guildID = preset["guild_id"];
+        const presetName = preset["preset_name"];
+        const options = JSON.parse(preset["game_options"]);
+        await Promise.all(
```
Brainicism, line 9: Let's add error handling here incase one of these fails and we need to do it manually

src/structures/guild_preference.ts
```diff
@@ -161,12 +197,17 @@ export default class GuildPreference {
...
         try {
-            await dbContext.kmq("game_option_presets")
-                .insert({
-                    guild_id: this.guildID,
-                    preset_name: presetName,
-                    game_options: JSON.stringify(this.gameOptions),
-                });
+            await Promise.all(
+                Object.entries(this.gameOptions).map(async (option) => {
+                    await dbContext.kmq("game_option_presets")
```
Brainicism, line 202: This should be a transaction

src/structures/guild_preference.ts
```diff
@@ -270,21 +325,25 @@ export default class GuildPreference {
...
 
-    /** @returns the current selected groups, if the groups option is active */
+    /** @returns the current selected groups by ID, if the groups option is active */
     getGroupIDs(): number[] {
         if (this.gameOptions.groups === null) return [];
         return this.gameOptions.groups.map((x) => x.id);
     }
 
+    /** @returns the current selected groups (name and ID), if the groups option is active */
+    getGroups(): MatchedArtist[] {
```
Brainicism, line 343: Why do we need this new function

src/structures/guild_preference.ts
```diff
@@ -321,6 +379,11 @@ export default class GuildPreference {
         return this.gameOptions.excludes.map((x) => x.id);
     }
 
+    /** @returns a list containing the excluded groups (name and ID) */
+    getExcludesGroups(): MatchedArtist[] {
```
Brainicism, line 383: same here

src/structures/guild_preference.ts
```diff
@@ -357,6 +419,11 @@ export default class GuildPreference {
         return this.gameOptions.includes.map((x) => x.id);
     }
 
+    /** @returns a list containing the included groups (name and ID) */
+    getIncludesGroups(): MatchedArtist[] {
```
Brainicism, line 423: same here

src/structures/guild_preference.ts
```diff
@@ -671,26 +725,35 @@ export default class GuildPreference {
...
      * Persists the current guild preference to the data store
      */
-    async updateGuildPreferences(updateGameSession: boolean) {
-        await dbContext.kmq("guild_preferences")
-            .where({ guild_id: this.guildID })
-            .update({ guild_preference: JSON.stringify(this) });
+    async updateGuildPreferences(updatedOptions: Array<{ name: string, value: GameOptionValue }>) {
+        await Promise.all(
+            updatedOptions.map(async (option) => {
+                await dbContext.kmq("game_options")
```
Brainicism, line 737: This should be a transaction

Brainicism (changes requested):

src/structures/guild_preference.ts
taahamahdi: I decided to replace all direct accesses of `this.gameOptions` with the relevant getter methods, similar to how I replaced directly setting member variables with setters in the reset methods (for [less points of failure](https://github.com/Brainicism/KMQ_Discord/pull/754/files#diff-6a5f3a950d16f8e31e2693b67c9ade56a23da2ebec004a48d0261571653f17edL531)). However, there weren't any equivalent functions for groups, includes, or excludes, so I decided to add them for consistency. However, I concede that it was mainly just a style choice on my end, and I'll happily revert them if you want.

Brainicism: Don't see a reason to expose this as a method

src/scripts/json-presets-to-new-format.ts
```diff
@@ -0,0 +1,37 @@
...
+const logger = _logger("json-presets-to-new-format");
+
+async function exportJsonPresetsToNewTable(db: DatabaseContext): Promise<void> {
+    const jsonPresets = await db.kmq("game_option_presets_json").select("*");
+    await Promise.all(jsonPresets.map(async (preset) => {
+        const guildID = preset["guild_id"];
+        const presetName = preset["preset_name"];
+        const options = JSON.parse(preset["game_options"]);
+        try {
+            await Promise.all(
```
Brainicism, line 13: This will throw an exception if _any_ of the promises fail. Should use `Promise.allSettled` and log the failed ones specifically

src/structures/guild_preference.ts
```diff
@@ -671,26 +709,40 @@ export default class GuildPreference {
...
+        await dbContext.kmq.transaction(async (trx) => {
+            await dbContext.kmq("game_options")
+                .insert(updatedOptions)
+                .onConflict(["guild_id", "option_name"])
+                .merge()
+                .transacting(trx);
+        });
         const gameSession = state.gameSessions[this.guildID];
-        if (gameSession && updateGameSession) {
+        if (gameSession) {
```
Brainicism, line 733: Why was the `updateGameSession` removed here?

Brainicism (changes requested):

src/scripts/json-presets-to-new-format.ts
taahamahdi: For a preset, if any of its option insertions fail, it should stop inserting without attempting to insert the remaining options

src/structures/guild_preference.ts
taahamahdi: Previously we called this function with `updateGameSession = false` in only one place, `fromGuild`. However, since fromGuild is passed with whatever is in the database already, it doesn't need to call `updateGuildPreferences`. For every remaining call of `updateGuildPreferences`, an active game session would need to be updated

src/scripts/json-presets-to-new-format.ts
Brainicism: This should be a transaction then

Brainicism (approved):

Brainicism:
Running E2E workflow: https://github.com/Brainicism/KMQ_Discord/actions/runs/868006055

PR #762, opened 2021-05-22
Fix bug where GameSession would be in stuck state

Brainicism:


src/config/knexfile_kmq.js
```diff
@@ -11,6 +11,6 @@ module.exports = {
     max: 10
   },
   migrations: {
-    directory: resolve(__dirname, "../migrations")
+    directory: resolve(__dirname, "../../src/migrations")
```
taahamahdi, line 14: This exits and re-enters the src directory

src/structures/game_session.ts
```diff
@@ -252,6 +252,7 @@ export default class GameSession {
      */
     endSession = async (): Promise<void> => {
         this.finished = true;
+        deleteGameSession(this.guildID);
```
taahamahdi, line 255: Does deleting the game session this early interfere with any accesses to it in the rest of this function?

taahamahdi (changes requested):

src/structures/game_session.ts
Brainicism: No, it doesn't disappear until it's garbage collected

taahamahdi (approved):

PR #783, opened 2021-06-05
Add publish_date_overrides table + trigger on available_songs

Brainicism:


sql/procedures/create_override_publish_date_procedure.sql
```diff
@@ -0,0 +1,25 @@
...
+ON publish_date_overrides
+FOR EACH ROW
+CALL OverridePublishDates() //
+
+DROP TRIGGER IF EXISTS insert_publish_overrides_trigger //
+CREATE TRIGGER insert_publish_overrides_trigger
+AFTER INSERT
+ON publish_date_overrides
+FOR EACH ROW
+CALL OverridePublishDates() //
```
taahamahdi, line 23: Not too sure of how triggers work, but is it possible to only call this on the inserted entry instead of each row (if that's not what it's already doing?)

taahamahdi (changes requested):

sql/procedures/create_override_publish_date_procedure.sql
Brainicism: It's possible to optimize this, but we'd need to have two separate procedures for updating all overrides and updating a single one. But it's not a very expensive call so didn't think it was worth doing

taahamahdi: Got it, thanks for clarifying

taahamahdi (approved):

PR #788, opened 2021-06-05
Remove avatar icons, add replies where appropriate

Brainicism:


src/commands/game_commands/profile.ts
```diff
@@ -59,16 +60,26 @@ export default class ProfileCommand implements BaseCommand {
...
         priority: 50,
     };
 
-    async call({ message }: CommandArgs) {
+    async call({ message, parsedMessage }: CommandArgs) {
         let requestedPlayer: Eris.User;
-        if (message.mentions.length === 1) {
+        if (parsedMessage.components.length === 0) {
+            requestedPlayer = message.author;
+        } else if (message.mentions.length === 1) {
```
taahamahdi, line 75: What happens if a user mentions more than 1 player? Will line 78 give an error or continue silently to the "profile not found" branch?

taahamahdi (approved):
Lgtm with question

taahamahdi (approved):

PR #794, opened 2021-06-11
Lower beginning cutoff to 2000

taahamahdi:


src/commands/game_options/cutoff.ts
```diff
@@ -6,7 +6,7 @@ import { GameOption } from "../../types";
 import MessageContext from "../../structures/message_context";
 
 const logger = _logger("cutoff");
-export const DEFAULT_BEGINNING_SEARCH_YEAR = 2005;
+export const DEFAULT_BEGINNING_SEARCH_YEAR = 2000;
```
Brainicism, line 9: Should be 1995, maybe 1990

taahamahdi:
Making a branch with generalized name and doing 1990

PR #854, opened 2021-07-17
Read bonus users from database instead of global state

Brainicism:
Using global state will lead to inconsistencies when moving to clustered solution. Read time from database should also be negligible, entire index should fit in cache

### DIFF
--- a/src/structures/game_session.ts
+++ b/src/structures/game_session.ts
@@ -255,7 +255,7 @@ export default class GameSession {
         }
         this.finished = true;
         deleteGameSession(this.guildID);
-        this.endRound({ correct: false }, await getGuildPreference(this.guildID));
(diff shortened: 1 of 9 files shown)


src/structures/game_session.ts
```diff
@@ -168,7 +168,7 @@ export default class GameSession {
      * @param guildPreference - The GuildPreference
      * @param messageContext - An object containing relevant parts of Eris.Message
      */
-    endRound(guessResult: GuessResult, guildPreference: GuildPreference, messageContext?: MessageContext) {
+    async endRound(guessResult: GuessResult, guildPreference: GuildPreference, messageContext?: MessageContext) {
```
taahamahdi, line 171: Should await every call to this now I think

taahamahdi (changes requested):

taahamahdi (approved):

PR #865, opened 2021-07-24
Implement importing/exporting presets

taahamahdi:


src/commands/game_commands/preset.ts
```diff
@@ -142,16 +170,65 @@ export default class PresetCommand implements BaseCommand {
...
+            .where("option_value", "=", presetUUID)
+            .first();
+
+        const presetOptions = await dbContext.kmq("game_option_presets")
+            .select(["option_name", "option_value"])
+            .where("guild_id", "=", existingPresetIdentifiers["guild_id"])
+            .andWhere("preset_name", "=", existingPresetIdentifiers["preset_name"]);
+
+        for (const option of presetOptions.filter((x) => x["option_name"] !== "uuid")) {
+            await dbContext.kmq("game_option_presets")
```
Brainicism, line 213: This should be a transaction

src/scripts/generate-preset-uuids.ts
```diff
@@ -0,0 +1,33 @@
...
+    const db = getNewConnection();
+    const presetsWithUUID = await db.kmq("game_option_presets")
+        .select(["guild_id", "preset_name"])
+        .distinct("guild_id", "preset_name")
+        .where("option_name", "=", "uuid");
+
+    const presetsWithoutUUID = await db.kmq("game_option_presets")
+        .select(["guild_id", "preset_name"])
+        .distinct("guild_id", "preset_name")
+        .whereNotIn(["guild_id", "preset_name"], presetsWithUUID.map((pair) => [pair["guild_id"], pair["preset_name"]]));
```
Brainicism, line 17: .whereNull?

src/structures/guild_preference.ts
```diff
@@ -185,22 +186,29 @@ export default class GuildPreference {
...
-     * @returns whether a preset was deleted
+     * @returns the old UUID if the deletion was successful and the empty string otherwise
      */
-    async deletePreset(presetName: string): Promise<boolean> {
-        const result = await dbContext.kmq("game_option_presets")
+    async deletePreset(presetName: string): Promise<string> {
+        const presetUUID = this.getPresetUUID(presetName);
+
+        if (!presetUUID) {
+            return "";
```
Brainicism, line 195: return null

src/structures/guild_preference.ts
```diff
@@ -228,7 +243,8 @@ export default class GuildPreference {
         const preset: { [x: string]: any } = (await dbContext.kmq("game_option_presets")
             .select(["option_name", "option_value"])
             .where("guild_id", "=", this.guildID)
-            .andWhere("preset_name", "=", presetName))
+            .andWhere("preset_name", "=", presetName)
+            .andWhere("option_name", "!=", "uuid"))
```
Brainicism, line 247: Don't even need to do this check, the uuid field will just go away when we cast to GameOption

src/structures/guild_preference.ts
```diff
@@ -254,6 +270,25 @@ export default class GuildPreference {
...
+    async getPresetUUID(presetName: string): Promise<string> {
+        const presetID = await dbContext.kmq("game_option_presets")
+            .select(["option_value"])
+            .where("guild_id", "=", this.guildID)
+            .andWhere("preset_name", "=", presetName)
+            .andWhere("option_name", "=", "uuid")
+            .first();
+
+        if (!presetID) {
+            return "";
```
Brainicism, line 286: Return null

Brainicism (changes requested):

src/structures/guild_preference.ts
taahamahdi: We still need to do this check because we perform a JSON.parse on each row and the UUID is stored normally

Brainicism: It'll get validated out in `validateGameOptions`

src/commands/game_commands/preset.ts
```diff
@@ -59,6 +63,14 @@ export default class PresetCommand implements BaseCommand {
                 example: "`,preset delete [preset_name]`",
                 explanation: "Deletes the mentioned preset",
             },
+            {
+                example: "`,preset export [preset_name]`",
+                explanation: "Returns a unique identifier associated with the mentioned preset",
```
Brainicism, line 68: Returns a unique identifier that can be used to load/import the preset

src/commands/game_commands/preset.ts
```diff
@@ -72,8 +84,15 @@ export default class PresetCommand implements BaseCommand {
             return;
         }
 
-        const presetName = parsedMessage.components[1];
+        const presetName = parsedMessage.components[presetAction !== PresetAction.IMPORT ? 1 : 2];
+        const presetUUID = parsedMessage.components[1];
         if (!presetName) {
+            if (presetAction === PresetAction.IMPORT && !presetUUID) {
```
Brainicism, line 90: Can we move this and `presetUUID` into the case statement

src/commands/game_commands/preset.ts
```diff
@@ -109,7 +134,26 @@ export default class PresetCommand implements BaseCommand {
...
 
     async loadPreset(presetName: string, guildPreference: GuildPreference, messageContext: MessageContext) {
-        const loadResult = await guildPreference.loadPreset(presetName);
+        let guildID = messageContext.guildID;
+        if (presetName.startsWith("KMQ-")) {
+            // User is loading a preset via UUID
+            const presetUUID = presetName;
+            const existingPresetIdentifiers = await dbContext.kmq("game_option_presets")
+                .select(["guild_id", "preset_name"])
+                .where("option_value", "=", JSON.stringify(presetUUID))
```
Brainicism, line 143: option_name = uuid?

src/commands/game_commands/preset.ts
```diff
@@ -142,22 +191,88 @@ export default class PresetCommand implements BaseCommand {
...
+    async importPreset(presetUUID: string, presetName: string, guildPreference: GuildPreference, messageContext: MessageContext) {
+        if ((await guildPreference.listPresets()).includes(presetName)) {
+            sendErrorMessage(messageContext, { title: "Preset Import Failed", description: `Preset \`${presetName}\` already exists. You can delete the old one with \`${process.env.BOT_PREFIX}preset delete ${presetName}\`.`, thumbnailUrl: KmqImages.DEAD });
+            logger.warn(`${getDebugLogHeader(messageContext)} | Preset import failed; '${presetName}' already exists`);
+            return;
+        }
+
+        const existingPresetIdentifiers = await dbContext.kmq("game_option_presets")
+            .select(["guild_id", "preset_name"])
+            .where("option_value", "=", JSON.stringify(presetUUID))
```
Brainicism, line 225: option_name = uuid

src/commands/game_commands/preset.ts
```diff
@@ -142,22 +191,88 @@ export default class PresetCommand implements BaseCommand {
...
+    }
+
+    async importPreset(presetUUID: string, presetName: string, guildPreference: GuildPreference, messageContext: MessageContext) {
+        if ((await guildPreference.listPresets()).includes(presetName)) {
+            sendErrorMessage(messageContext, { title: "Preset Import Failed", description: `Preset \`${presetName}\` already exists. You can delete the old one with \`${process.env.BOT_PREFIX}preset delete ${presetName}\`.`, thumbnailUrl: KmqImages.DEAD });
+            logger.warn(`${getDebugLogHeader(messageContext)} | Preset import failed; '${presetName}' already exists`);
+            return;
+        }
+
+        const existingPresetIdentifiers = await dbContext.kmq("game_option_presets")
```
Brainicism, line 223: is this supposed to be plural

src/commands/game_commands/preset.ts
```diff
@@ -109,7 +134,26 @@ export default class PresetCommand implements BaseCommand {
     }
 
     async loadPreset(presetName: string, guildPreference: GuildPreference, messageContext: MessageContext) {
-        const loadResult = await guildPreference.loadPreset(presetName);
+        let guildID = messageContext.guildID;
+        if (presetName.startsWith("KMQ-")) {
+            // User is loading a preset via UUID
+            const presetUUID = presetName;
+            const existingPresetIdentifiers = await dbContext.kmq("game_option_presets")
```
Brainicism, line 141: is this supposed to be plural

src/commands/game_commands/preset.ts
```diff
@@ -142,22 +191,88 @@ export default class PresetCommand implements BaseCommand {
...
+                }));
+
+            await dbContext.kmq("game_option_presets")
+                .insert(preset)
+                .onConflict(["guild_id", "preset_name", "option_name"])
+                .merge()
+                .transacting(trx);
+
+            await dbContext.kmq("game_option_presets")
+                .insert({
```
Brainicism, line 256: Can we just add the this UUID object to `preset` and only insert once?

Brainicism (changes requested):

src/commands/game_commands/preset.ts
taahamahdi: We need to find the preset with the matching uuid, since all presets have a uuid option_name doing that won't work

src/commands/game_commands/preset.ts
taahamahdi: I decided to call it that since the two identifiers are guild_id and preset_name

src/commands/game_commands/preset.ts
```diff
@@ -72,10 +84,14 @@ export default class PresetCommand implements BaseCommand {
...
 
-        const presetName = parsedMessage.components[1];
-        if (!presetName) {
+        const missingPresetMessage = (() => {
             sendErrorMessage(messageContext, { title: "Preset Name Missing", description: "You must specify a preset name.", thumbnailUrl: KmqImages.NOT_IMPRESSED });
             logger.warn(`${getDebugLogHeader(message)} | Preset name not specified`);
+        });
+
+        const presetName = parsedMessage.components[presetAction !== PresetAction.IMPORT ? 1 : 2];
+        if (presetAction !== PresetAction.IMPORT && !presetName) {
```
Brainicism, line 93: Why do we still need this if it's already checked in the case statement

taahamahdi: If it were removed, in the case where neither the UUID nor the preset name are entered, it will only give the error message for the missing preset name, even though the UUID is the first argument

Brainicism (approved):

PR #898, opened 2021-08-07
Use formatted number in leaderboard

taahamahdi:
* Also remove code formatting from leaderboard footer

### DIFF
--- a/src/commands/game_commands/leaderboard.ts
+++ b/src/commands/game_commands/leaderboard.ts
@@ -27,8 +27,8 @@ enum LeaderboardAction {
 }
 
 const leaderboardQuotes = [
-    `Want your name to show up on the leaderboard? See \`${process.env.BOT_PREFIX}help leaderboard\``,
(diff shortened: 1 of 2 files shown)


src/commands/game_commands/leaderboard.ts
```diff
@@ -27,8 +27,8 @@ enum LeaderboardAction {
 }
 
 const leaderboardQuotes = [
-    `Want your name to show up on the leaderboard? See \`${process.env.BOT_PREFIX}help leaderboard\``,
-    `Want to see the next page of the leaderboard? See \`${process.env.BOT_PREFIX}help leaderboard\``,
+    `Want your name to show up on the leaderboard? See ${process.env.BOT_PREFIX}help leaderboard`,
```
Brainicism, line 30: Why get rid of code line here?

Brainicism (changes requested):

src/commands/game_commands/leaderboard.ts
taahamahdi: They're footers so the code line doesn't work

Brainicism (approved):

PR #941, opened 2021-08-18
Fixed bug where multiple choice wouldn't decrement lives or end game in elimination

### LINKED ISSUE #937: Elimination doesn't work with ,answer
Guessing incorrectly doesn't decrement lives

taahamahdi:
Closes #937.

### DIFF
--- a/src/structures/game_session.ts
+++ b/src/structures/game_session.ts
@@ -263,11 +263,6 @@ export default class GameSession {
             this.endSession();
         } else if (this.scoreboard.gameFinished(guildPreference)) {
             this.endSession();
-        } else if (this.gameType === GameType.ELIMINATION) {
-            const eliminationScoreboard = this.scoreboard as EliminationScoreboard;
-            if (eliminationScoreboard.gameFinished()) {
-                this.endSession();
-            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/structures/game_session.ts
```diff
@@ -259,6 +263,11 @@ export default class GameSession {
             this.endSession();
         } else if (this.scoreboard.gameFinished(guildPreference)) {
             this.endSession();
+        } else if (this.gameType === GameType.ELIMINATION) {
+            const eliminationScoreboard = this.scoreboard as EliminationScoreboard;
+            if (eliminationScoreboard.gameFinished()) {
```
Brainicism, line 268: Isn't this the same as line 265

Brainicism (changes requested):

src/structures/game_session.ts
taahamahdi: Since the subclass and superclass have different signatures I think I have to do this. Should I add guildPreferences as an unused argument to elimination scoreboard and remove these lines?

Brainicism: Let's do that

taahamahdi: Nevermind I'm wrong, I just tested it like you suggested and it works

Brainicism (approved):

PR #949, opened 2021-08-19
5% bonus EXP when using a special mode

taahamahdi:


src/structures/game_session.ts
```diff
@@ -1052,6 +1052,10 @@ export default class GameSession {
             }
         }
 
+        if (guildPreference.gameOptions.specialType) {
+            expModifier *= 1.05;
```
Brainicism, line 1056: Is 5% too low? Should the modifiers depend on which special type you're playing?

taahamahdi: 5% might be too low, you're right. In terms of difficulty, maybe it goes reverse > faster > fast > slow?

taahamahdi: I could scale it like so: 

Slow: 5%

Fast: 10%

Faster: 15%

Reverse: 20%

PR #970, opened 2021-08-28
Allow profile via user command interaction

taahamahdi:


src/commands/game_commands/profile.ts
```diff
@@ -50,6 +50,115 @@ export function getRankNameByLevel(level: number): string {
...
+    const gamesPlayed = playerStats["games_played"];
+    const firstPlayDateString = friendlyFormattedDate(new Date(playerStats["first_play"]));
+    const lastActiveDateString = friendlyFormattedDate(new Date(playerStats["last_active"]));
+
+    const totalPlayers = (await dbContext.kmq("player_stats")
+        .count("* as count")
+        .where("exp", ">", "0")
+        .first())["count"] as number;
+
+    const { exp, level } = (await dbContext.kmq("player_stats")
```
Brainicism, line 73: Isn't the query on line 54 the same as this

taahamahdi: Yeah it is, I just copied over the existing implementation into this function. Should I combine the queries?

Brainicism: Yes

src/events/client/interactionCreate.ts
```diff
@@ -131,18 +132,51 @@ export default async function interactionCreateHandler(interaction: Eris.PingInt
...
-        // Bookmarking songs
-        const gameSession = state.gameSessions[interaction.guildID];
-        const song = gameSession?.getSongFromMessageID(interaction.data.target_id);
-        if ((!gameSession || !song) && withinInteractionInterval(interaction)) {
-            await tryCreateErrorMessage(interaction, !gameSession ? "You can only bookmark songs during a game." : `You can only bookmark songs recently played in the last ${BOOKMARK_MESSAGE_SIZE} rounds. You must bookmark the message sent by the bot containing the song.`);
-            return;
-        }
+        if (interaction.data.type === Eris.Constants.ApplicationCommandTypes.USER) {
+            // Profile
+            const userID = interaction.data.target_id;
```
Brainicism, line 137: Can we refactor this file to not include business logic? 
1. Move the multiple choice selection logic to game_session
2. Move this logic to `,profile`
3. Move bookmark logic to game_session  

Also, how do we differentiate between interactions of the same type, but different commands?

Brainicism (changes requested):

src/events/client/interactionCreate.ts
taahamahdi: > Also, how do we differentiate between interactions of the same type, but different commands?

`interaction.data.name`

Brainicism (approved):

PR #990, opened 2021-09-04
Null check for gameround in guess timeout function

Brainicism:


src/structures/game_session.ts
```diff
@@ -627,7 +627,7 @@ export default class GameSession {
 
         const time = guildPreference.gameOptions.guessTimeout;
         this.guessTimeoutFunc = setTimeout(async () => {
-            if (this.finished || this.gameRound.finished) return;
+            if (this.finished || this.gameRound?.finished) return;
```
taahamahdi, line 630: If there is no game round then this will return false, so the early return won't happen. Add a separate check for that instead

taahamahdi (changes requested):

taahamahdi (approved):

PR #992, opened 2021-09-05
Add competition mode

taahamahdi:


src/commands/game_commands/play.ts
```diff
@@ -17,6 +17,11 @@ import KmqMember from "../../structures/kmq_member";
 const logger = new IPCLogger("play");
 const DEFAULT_LIVES = 10;
 
+export const COMPETITION_MODERATOR_IDS = [
```
Brainicism, line 20: Let's not hardcode IDs into master

src/commands/game_commands/play.ts
```diff
@@ -148,7 +154,7 @@ export default class PlayCommand implements BaseCommand {
                 const startTitle = `\`${prefix}join\` the game and start it with \`${prefix}begin\`!`;
                 const gameInstructions = `Type \`${prefix}join\` to play in the upcoming elimination game. Once all have joined, ${getMention(gameOwner.id)} must send \`${prefix}begin\` to start the game. Everyone begins with \`${lives}\` lives.`;
 
-                gameSession = new GameSession(textChannel.id, voiceChannel.id, textChannel.guild.id, gameOwner, GameType.ELIMINATION, lives);
+                gameSession = new GameSession(textChannel.id, voiceChannel.id, textChannel.guild.id, gameOwner, GameType.ELIMINATION, isCompetitionMode, lives);
```
Brainicism, line 157: I don't think it's possible to be both `competition` and `elimination | teams`?

src/commands/game_commands/play.ts
```diff
@@ -170,7 +176,12 @@ export default class PlayCommand implements BaseCommand {
                     sendErrorMessage(messageContext, { title: ignoringOldGameTypeTitle, description: oldGameTypeInstructions, thumbnailUrl: KmqImages.DEAD });
                 }
 
-                gameSession = new GameSession(textChannel.id, voiceChannel.id, textChannel.guild.id, gameOwner, GameType.CLASSIC);
+                if (isCompetitionMode && !COMPETITION_MODERATOR_IDS.includes(message.author.id)) {
```
Brainicism, line 179: What if we just restrict the command to the debug channel, instead of checking for moderator id

src/commands/game_commands/play.ts
Brainicism: Any reason we don't make `GameType.COMPETITION` instead of new variable to keep track?

Brainicism (changes requested):

src/commands/game_commands/play.ts
taahamahdi: The debug channel will be different from the one that hosts the competition

src/commands/game_commands/play.ts
taahamahdi: > Any reason we don't make `GameType.COMPETITION` instead of new variable to keep track?

I can do this, but I will also need to refactor any CLASSIC checks in code and include competition there.

Once we enable argument validation here, if a user enters an invalid game type, the expected values shown will include competition even though it can't be used.

taahamahdi: > I don't think it's possible to be both `competition` and `elimination | teams`?

You're right. I put the variable there to show what the argument being passed was (instead of just writing false), but it can only be false with those game modes. If I refactor to make COMPETITION a GameType, I'll remove this

src/commands/game_commands/play.ts
taahamahdi: I can pass them as environment variables

src/events/client/messageCreate.ts
```diff
@@ -52,6 +53,14 @@ export default async function messageCreateHandler(message: Eris.Message) {
                 if (!preCheckResult) return;
             }
 
+            const gameSession = gameSessions[message.guildID];
+            if (gameSession && gameSession.gameType === GameType.COMPETITION) {
+                if (!COMPETITION_MODERATOR_IDS.includes(message.author.id) && !["skip", "score", "scoreboard", "options", "help", "news", "stats", "s", "profile", "leaderboard"].includes(parsedMessage.action)) {
```
Brainicism, line 58: You can add a new prerunCheck for all of these commands instead of hardcoding list here

src/command_prechecks.ts
```diff
@@ -28,21 +29,39 @@ export function inGameCommandPrecheck(message: GuildTextableMessage, gameSession
...
+
+export async function competitionPrecheck(message: GuildTextableMessage, gameSession: GameSession, errorMessage?: string): Promise<boolean> {
+    if (!gameSession || gameSession.gameType !== GameType.COMPETITION) {
+        return true;
+    }
+
+    const isModerator = (await dbContext.kmq("competition_moderators").select("user_id")
+        .where("guild_id", "=", gameSession.guildID)
+        .andWhere("user_id", "=", message.author.id)
+        .first()) ?? false;
```
Brainicism, line 59: ?? null

src/commands/game_commands/begin.ts
```diff
@@ -8,12 +8,15 @@ import { IPCLogger } from "../../logger";
...
+import { competitionPrecheck } from "../../command_prechecks";
 
 const logger = new IPCLogger("begin");
 
 export default class BeginCommand implements BaseCommand {
+    preRunChecks = [{ checkFn: competitionPrecheck }];
+
     canStart(gameSession: GameSession, authorID: string, messageContext: MessageContext): boolean {
-        if (!gameSession || gameSession.gameType === GameType.CLASSIC) {
+        if (!gameSession || [GameType.CLASSIC, GameType.COMPETITION].includes(gameSession.gameType)) {
```
Brainicism, line 19: Does it make more sense for this to be `gameSession.gameType !== GameType.TEAMS`

src/commands/game_commands/join.ts
```diff
@@ -9,13 +9,16 @@ import { bold } from "../../helpers/utils";
...
 
 export default class JoinCommand implements BaseCommand {
+    preRunChecks = [{ checkFn: competitionPrecheck }];
+
     aliases = ["j"];
 
     call = async ({ message, gameSessions, parsedMessage }: CommandArgs) => {
         const gameSession = gameSessions[message.guildID];
-        if (!gameSession || gameSession.gameType === GameType.CLASSIC) {
+        if (!gameSession || [GameType.CLASSIC, GameType.COMPETITION].includes(gameSession.gameType)) {
```
Brainicism, line 21: same here

src/commands/game_commands/play.ts
```diff
@@ -131,7 +134,7 @@ export default class PlayCommand implements BaseCommand {
         const prefix = process.env.BOT_PREFIX;
 
         if (!gameSessions[message.guildID] || !gameSessions[message.guildID].sessionInitialized) {
-            // (1) No game session exists yet (create MC, ELIMINATION, TEAMS, or CLASSIC game), or
+            // (1) No game session exists yet (create MC, ELIMINATION, TEAMS, CLASSIC, or COMPETITION game), or
```
Brainicism, line 137: old, but why is MC mentioned here

src/commands/game_commands/play.ts
```diff
@@ -170,7 +173,18 @@ export default class PlayCommand implements BaseCommand {
                     sendErrorMessage(messageContext, { title: ignoringOldGameTypeTitle, description: oldGameTypeInstructions, thumbnailUrl: KmqImages.DEAD });
                 }
 
-                gameSession = new GameSession(textChannel.id, voiceChannel.id, textChannel.guild.id, gameOwner, GameType.CLASSIC);
+                const isCompetitionMode = parsedMessage.components.length >= 1 && parsedMessage.components[0].toLowerCase() === "competition";
+                const isModerator = (await dbContext.kmq("competition_moderators").select("user_id")
```
Brainicism, line 177: We don't need this check unless we're in competition mode

Brainicism (changes requested):

src/commands/game_commands/begin.ts
taahamahdi: I'd need to check that it's not elimination and it's not teams, but it does make more sense I think

Brainicism (approved):

PR #993, opened 2021-09-05
Convert command prechecks from abstract class to base command property

Brainicism:


src/command_prechecks.ts
```diff
@@ -0,0 +1,49 @@
...
+        logger.warn(`${getDebugLogHeader(message)} | User and bot are not in the same voice connection`);
+        sendErrorMessage(MessageContext.fromMessage(message), { title: "Wait...", description: "You must be in the same voice channel as the bot to use this command." });
+        return false;
+    }
+
+    return true;
+}
+
+export function debugServerPrecheck(message: GuildTextableMessage, _gameSession: GameSession): boolean {
+    const res = process.env.DEBUG_SERVER_ID === message.guildID;
```
taahamahdi, line 32: Nit: you can avoid the temporary variable here and in the function below

taahamahdi (approved):

src/command_prechecks.ts
Brainicism: Need the result of the variable to return the warning message if true

taahamahdi: Could return false if there's an error, and true if there isn't

PR #1003, opened 2021-09-13
Convert all arguments passed to lowercase

taahamahdi:
Since the user passes a team with ,join, it isn't converted to lowercase


Brainicism (approved):

Brainicism (changes requested):
What about preset names?

PR #1025, opened 2021-09-24
Implement per-user rate limiting

Brainicism:
Closes #1025

### DIFF
--- a/src/ratelimiter.ts
+++ b/src/ratelimiter.ts
@@ -1,0 +1,55 @@
+export default class RateLimiter {
+    public limitMap: { [userID: string]: Array<number> };
+    private queueCapacity: number;
+    private queueExpiryTime: number;
+
+    constructor(queueCapacity: number, queueExpiryTime: number) {
+        this.queueCapacity = queueCapacity;
+        this.queueExpiryTime = queueExpiryTime;
+        this.limitMap = {};
+    }
+
+    /**
+     * @param userID - the user's ID
+     * @returns the time remaining until the next request can be processed, in milliseconds
+     */
+    public timeRemaining(userID: string): number {
+        if (!(userID in this.limitMap)) return 0;
+        const previousRequestDates = this.limitMap[userID];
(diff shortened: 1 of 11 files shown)


src/ratelimiter.ts
```diff
@@ -0,0 +1,55 @@
+export default class RateLimiter {
```
taahamahdi, line 1: rate_limiter.ts, similar change to test

src/test/ci/ratelimiter.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe("multiple users", () => {
+    const userID2 = "234";
+    describe("user 1 hits limit, user 2 doesn't", () => {
+        it("should pass all requests until capacity for user 1, and fail else. user 2 should all pass", () => {
+            for (let i = 0; i < queueCapacity; i++) {
+                assert.strictEqual(rateLimiter.check(userID), true);
+            }
+
+            for (let i = 0; i < queueCapacity / 2; i++) {
```
taahamahdi, line 138: Could go up to `queueCapacity - 1` to be a bit more strict with the check

src/ratelimiter.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * @param userID - the user's ID
+     * @returns the time remaining until the next request can be processed, in milliseconds
+     */
+    public timeRemaining(userID: string): number {
+        if (!(userID in this.limitMap)) return 0;
+        const previousRequestDates = this.limitMap[userID];
+        return (previousRequestDates[0] + (this.queueExpiryTime * 1000)) - Date.now();
```
taahamahdi, line 19: Can this ever be negative? If it can, should return 0 in that case

taahamahdi (changes requested):

src/test/ci/ratelimiter.test.ts
Brainicism: This test tests that the requests from one user doesn't impact the requests from another user, intentionally did not fill it to capacity

taahamahdi (approved):

PR #1107, opened 2021-11-21
Add daily bonus artists

taahamahdi:


Brainicism (approved):

src/helpers/management_utils.ts
```diff
@@ -218,6 +218,19 @@ export async function reloadAliases() {
     logger.info("Reloaded alias data");
 }
 
+/** Reload bonus groups (same groups chosen on the same day) */
+export async function reloadBonusGroups() {
+    const bonusGroupCount = 10;
+    const date = new Date();
+    const artistNameQuery: string[] = (await dbContext.kmq("kpop_groups")
+        .select(["name"])
```
Brainicism, line 226: Should we avoid collabs?

Brainicism (changes requested):

src/helpers/management_utils.ts
taahamahdi: Yeah we should

Brainicism (approved):

PR #1114, opened 2021-11-21
Reuse song guess message with ,answer

### LINKED ISSUE #1097: Replace 'guess the song' embed with results in multiple choice mode

taahamahdi:
Closes #1097.

* Fix bug where ,end doesn't send end of round message

### DIFF
--- a/src/helpers/discord_utils.ts
+++ b/src/helpers/discord_utils.ts
@@ -475,31 +475,27 @@ export async function sendEndRoundMessage(messageContext: MessageContext,
     }
 
     const songAndArtist = bold(`"${gameRound.originalSongName}" - ${gameRound.artistName}`);
-    if (isMultipleChoiceMode) {
-        await gameRound.interactionMessage.edit({
-            embeds: [{
-                color,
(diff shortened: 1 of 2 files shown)


src/helpers/discord_utils.ts
```diff
@@ -475,6 +475,22 @@ export async function sendEndRoundMessage(messageContext: MessageContext,
     }
 
     const songAndArtist = bold(`"${gameRound.originalSongName}" - ${gameRound.artistName}`);
+    if (isMultipleChoiceMode) {
+        await gameRound.interactionMessage.edit({
+            embeds: [{
```
Brainicism, line 480: Re-use the embed in both cases

Brainicism (changes requested):

src/helpers/discord_utils.ts
taahamahdi: I don't think I can because of the differing thumbnailUrl and footerText fields

Brainicism (approved):

PR #1133, opened 2021-12-09
Leaderboard by songs guessed and games played

taahamahdi:


src/commands/game_commands/leaderboard.ts
```diff
@@ -28,6 +28,12 @@ const logger = new IPCLogger("leaderboard");
 export const ENTRIES_PER_PAGE = 10;
 
 export enum LeaderboardType {
+    EXP = "exp",
+    GAMES_PLAYED = "games_played",
```
Brainicism, line 32: Should these be hyphens?

src/commands/game_commands/leaderboard.ts
```diff
@@ -169,14 +182,50 @@ export default class LeaderboardCommand implements BaseCommand {
         if (parsedMessage.components.length === 1) {
             LeaderboardCommand.showLeaderboard(
                 message,
-                type ?? LeaderboardType.GLOBAL,
+                type ?? LeaderboardType.EXP,
```
Brainicism, line 185: Add these as default values to the function instead

Brainicism (changes requested):

src/commands/game_commands/leaderboard.ts
```diff
@@ -28,13 +28,19 @@ const logger = new IPCLogger("leaderboard");
 export const ENTRIES_PER_PAGE = 10;
 
 export enum LeaderboardType {
+    EXP = "exp",
+    GAMES_PLAYED = "gamesplayed",
```
Brainicism, line 32: Can this just be 'songs' and 'games'?

taahamahdi: I initially considered it, but decided against it because `game` is already an argument

Brainicism (approved):

src/commands/game_commands/leaderboard.ts
taahamahdi: We can alternatively separate this into three commands, and maintain the logic in this file

PR #1146, opened 2021-12-16
Store per-song skip, hint count, time played, and time to guess

taahamahdi:
* Rename `song_guess_count` to `song_metadata`

### DIFF
--- a/src/migrations/20211216025236_song_guess_count_to_song_metadata.js
+++ b/src/migrations/20211216025236_song_guess_count_to_song_metadata.js
@@ -1,0 +1,19 @@
+exports.up = function (knex) {
+    return knex.schema
+        .renameTable("song_guess_count", "song_metadata")
+        .table("song_metadata", (table) => {
+            table.integer("skip_count").defaultTo(0);
+            table.integer("hint_count").defaultTo(0);
+            table.integer("time_played_ms").defaultTo(0);
+        });
+};
+
+exports.down = function (knex) {
+    return knex.schema
+        .table("song_metadata", (table) => {
+            table.dropColumn("skip_count");
+            table.dropColumn("hint_count");
+            table.dropColumn("time_played_ms");
+        })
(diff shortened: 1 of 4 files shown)


src/structures/game_session.ts
```diff
@@ -215,6 +221,7 @@ export default class GameSession {
         this.gameRound = null;
 
         gameRound.interactionMarkAnswers(guessResult.correctGuessers?.length);
+        const guessSpeed = Date.now() - gameRound.startedAt;
```
Brainicism, line 224: If you move it out here, this means that skips count as well?

Brainicism: Should we rename this?

Brainicism (changes requested):

src/structures/game_session.ts
taahamahdi: Oops, I'll rename it to `timePlayed`

Brainicism (approved):

src/migrations/20211216025236_song_guess_count_to_song_metadata.js
```diff
@@ -0,0 +1,31 @@
+exports.up = function (knex) {
+    return knex.schema
+        .renameTable("song_guess_count", "song_metadata")
+        .table("song_metadata", (table) => {
+            table.renameColumn("correct_guesses", "correct_guesses_legacy");
```
Brainicism, line 5: Don't we need to undo this rename in the exports.down?

Brainicism (changes requested):

src/migrations/20211216025236_song_guess_count_to_song_metadata.js
taahamahdi: It falls back to the existing columns (correct_guesses and rounds_played will already exist)

Brainicism (approved):

PR #1173, opened 2022-01-12
Add eslint rules for i18n translation files

Brainicism:


.eslintrc.i18n.js
```diff
@@ -0,0 +1,13 @@
+const path = require("path");
+
+module.exports = {
+    extends: ["plugin:i18n-json/recommended"],
+    rules: {
+        "i18n-json/identical-keys": [
+            "warn",
+            {
+                filePath: path.resolve("./i18n/en.json"),
```
taahamahdi, line 9: Can you pass ko.json/all files in that directory?

taahamahdi (changes requested):

.eslintrc.i18n.js
Brainicism: This is the reference file to compare against

Brainicism: This is the reference file to compare against

taahamahdi (approved):

PR #1217, opened 2022-02-18
Per-client-ID game options

taahamahdi:
Also, adds `CLIENT_ID` to .env and replaces `state.client.user.id`


src/migrations/20220218013908_client_id_in_game_options.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.schema.table("game_options", (table) => {
+        table
+            .text("client_id")
+            .notNullable()
+            .defaultTo(process.env.BOT_CLIENT_ID);
```
Brainicism, line 10: Does the migration run with these environment variables loaded?

PR #1226, opened 2022-02-27
Refactor of participants, ,play elimination

taahamahdi:
* Create Player upon VC entry instead of on-demand or elimination game start
* Remove redundant `participants` in GameSession and use scoreboard to manage instead
* Start elimination game with current VC members instead, and don't require to ,join
* Remove `Scoreboard::isEmpty()` and replace previous checks using it with checking to see if `Scoreboard::firstPlace` is empty (they were relying on the side effect that there are no scoreboard entries when there is no one in first place)
* Show players with 0 points on scoreboard, but only if they are in VC
* Provide more accurate counts of active players for stats by checking players are currently in VC
* Add `EliminationPlayer::fromUserID()`
* Remove redundant `Player::getID()`/`getName()` (already `public readonly`)

### DIFF
--- a/src/structures/game_session.ts
+++ b/src/structures/game_session.ts
@@ -41,7 +41,7 @@ import {
 } from "../helpers/utils";
 import { state } from "../kmq_worker";
 import { IPCLogger } from "../logger";
-import { QueriedSong, PlayerRoundResult, GameType, EnvType } from "../types";
+import { QueriedSong, PlayerRoundResult, GameType } from "../types";
 import GameRound from "./game_round";
 import GuildPreference from "./guild_preference";
 import Scoreboard, { SuccessfulGuessResult } from "./scoreboard";
@@ -205,24 +205,16 @@ export default class GameSession {
             this.scoreboard = new Scoreboard();
         }
 
-        if (
-            ![EnvType.CI, EnvType.TEST].includes(
-                process.env.NODE_ENV as EnvType
-            )
-        ) {
-            getCurrentVoiceMembers(this.voiceChannelID)
-                .filter((x) => x.id !== process.env.BOT_CLIENT_ID)
-                .map((x) => x.id)
-                .map((x) =>
-                    this.scoreboard.addPlayer(
-                        [GameType.CLASSIC, GameType.COMPETITION].includes(
-                            this.gameType
-                        )
-                            ? Player.fromUserID(x)
-                            : EliminationPlayer.fromUserID(x, eliminationLives)
-                    )
-                );
-        }
+        getCurrentVoiceMembers(this.voiceChannelID)
+            .filter((x) => x.id !== process.env.BOT_CLIENT_ID)
+            .map((x) => x.id)
+            .map((x) =>
+                this.scoreboard.addPlayer(
+                    this.gameType === GameType.ELIMINATION
+                        ? EliminationPlayer.fromUserID(x, eliminationLives)
+                        : Player.fromUserID(x)
+                )
+            );
     }
 
     /**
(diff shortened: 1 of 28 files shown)


src/structures/game_session.ts
```diff
@@ -1187,6 +1187,22 @@ export default class GameSession {
      * @param inVC - Whether the player is currently in the voice channel
      */
     setPlayerInVC(userID: string, inVC: boolean): void {
+        if (
+            inVC &&
+            !this.scoreboard.getPlayerIDs().includes(userID) &&
+            this.gameType !== GameType.TEAMS
+        ) {
+            this.scoreboard.addPlayer(
+                [GameType.CLASSIC, GameType.COMPETITION].includes(this.gameType)
```
Brainicism, line 1196: would it not making more sense to check `this.gameType == elimination`?

src/test/ci/begin.test.ts
```diff
@@ -88,7 +53,7 @@ describe("begin command", () => {
                     assert.strictEqual(
                         BeginCommand.canStart(
                             gameSession,
-                            "1231",
+
```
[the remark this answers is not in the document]
taahamahdi: Whoops, thanks

src/structures/game_session.ts
taahamahdi: You're right, thanks :]

src/structures/scoreboard.ts
```diff
@@ -296,12 +289,18 @@ export default class Scoreboard {
...
      */
-    static getRanking(players: { [userID: string]: Player }): Array<string> {
-        return Object.entries(players)
-            .sort((a, b) => b[1].getScore() - a[1].getScore())
-            .map((x) => x[0]);
+    getRanking(): { [score: number]: number } {
+        const rankingToScore = {};
+        const sortedUniqueScores = [
+            ...new Set(Object.values(this.players).map((x) => x.getScore())),
+        ].sort((a, b) => b - a);
```
taahamahdi, line 298: Not sure if the sort is needed here anymore

src/test/ci/scoreboard.test.ts
```diff
@@ -99,22 +101,102 @@ describe("score/exp updating", () => {
     describe("position changes", () => {
         it("should return the correct ranking of every player", async () => {
             const players = {
-                "12345": new Player("", "12345", "", 2),
+                ohmiID: new Player("", "ohmiID", "", 2),
```
Brainicism, line 104: Add tests for every player having different score, every player having same score

src/structures/player.ts
```diff
@@ -42,24 +45,25 @@ export default class Player {
         this.avatarURL = avatarURL;
         this.expGain = 0;
         this.firstGameOfTheDay = firstGameOfTheDay;
+        this.previousRoundRanking = -1;
```
Brainicism, line 48: Why not `null`?

src/structures/scoreboard.ts
Brainicism: Yes required

src/structures/scoreboard.ts
```diff
@@ -296,12 +289,18 @@ export default class Scoreboard {
...
     /**
-     * @param players - user IDs mapped to Players
-     * @returns players IDs, sorted by rank in the game
+     * @returns a mapping of player scores to ranking
      */
-    static getRanking(players: { [userID: string]: Player }): Array<string> {
-        return Object.entries(players)
-            .sort((a, b) => b[1].getScore() - a[1].getScore())
-            .map((x) => x[0]);
+    getRanking(): { [score: number]: number } {
```
Brainicism, line 294: getScoreToRankingMap?

Brainicism (changes requested):

Brainicism (approved):

PR #1237, opened 2022-03-06
Remove Discord discriminator, use bold score

taahamahdi:


src/structures/player.ts
```diff
@@ -2,12 +2,14 @@ import {
...
+import { getMention } from "../helpers/discord_utils";
 import { bold } from "../helpers/utils";
 import { state } from "../kmq_worker";
 
 export default class Player {
-    /** The Discord tag of the player, of the format "Player#1234" */
+    /** The Discord username of the player sans discriminator,
+     * i.e. "Player" when the player's user tag is "Player#1234"
+     */
     public readonly name: string;
```
Brainicism, line 13: Rename to username?

Brainicism (changes requested):

src/structures/player.ts
taahamahdi: It makes sense here but it doesn't in `team`

Brainicism (approved):

PR #1239, opened 2022-03-07
Fix bug where first game of day bonus wasn't applied

taahamahdi:


src/structures/game_session.ts
```diff
@@ -1202,7 +1208,7 @@ export default class GameSession {
         this.scoreboard
             .getPlayerIDs()
             .filter((x) => !currentVoiceMembers.includes(x))
-            .map((x) => this.setPlayerInVC(x, false));
+            .map(async (x) => this.setPlayerInVC(x, false));
```
Brainicism, line 1211: s

Brainicism, line 1211:
```suggestion
            .map(async (x) => await this.setPlayerInVC(x, false));
```

Brainicism (changes requested):

Brainicism (approved):

PR #1250, opened 2022-03-31
Add ,music command

taahamahdi:
I separated some logic out of GameSession into a new class, Session. The new MusicSession class inherits from Session. Commands no longer receive gameSessions passed as an argument. Accessing a session should be done with `Session.getSession()` instead of accessing state directly.

The music command is pretty barebones right now. The only obvious frontend change is that it sends the round message at the start of the round instead of at the end. I'd like to add a ~~text-based duration slider that auto-updates the song message every 5 seconds~~ command to the song's progress. I'd also like to add a skip button and a bookmark button. The options message should also be updated to ~~only show relevant options when a music session is active~~ cross out irrelevant options, and seek should be automatically forced to beginning.

### DIFF
--- a/src/commands/game_commands/forceskip.ts
+++ b/src/commands/game_commands/forceskip.ts
@@ -77,10 +77,7 @@ export default class ForceSkipCommand implements BaseCommand {
             MessageContext.fromMessage(message),
             {
                 color: EMBED_SUCCESS_COLOR,
-                title: state.localizer.translate(
-                    message.guildID,
-                    "command.skip.success.title"
-                ),
+                title: state.localizer.translate(message.guildID, "misc.skip"),
                 description: state.localizer.translate(
                     message.guildID,
                     "command.forceskip.description"
--- a/src/commands/game_commands/forceskip.ts
+++ b/src/commands/game_commands/forceskip.ts
@@ -13,6 +13,7 @@ import MessageContext from "../../structures/message_context";
 import { KmqImages } from "../../constants";
 import CommandPrechecks from "../../command_prechecks";
 import { state } from "../../kmq_worker";
+import Session from "../../structures/session";
 
 const logger = new IPCLogger("forceskip");
 
@@ -20,7 +21,7 @@ export default class ForceSkipCommand implements BaseCommand {
     aliases = ["fskip", "fs"];
 
     preRunChecks = [
-        { checkFn: CommandPrechecks.inGameCommandPrecheck },
+        { checkFn: CommandPrechecks.inSessionCommandPrecheck },
         { checkFn: CommandPrechecks.competitionPrecheck },
     ];
 
@@ -35,32 +36,29 @@ export default class ForceSkipCommand implements BaseCommand {
         priority: 1009,
     });
 
-    call = async ({ gameSessions, message }: CommandArgs): Promise<void> => {
-        const guildPreference = await getGuildPreference(message.guildID);
-        const gameSession = gameSessions[message.guildID];
-        if (
-            !gameSession ||
-            !gameSession.round ||
-            !areUserAndBotInSameVoiceChannel(message)
-        ) {
+    call = async ({ message }: CommandArgs): Promise<void> => {
+        if (!areUserAndBotInSameVoiceChannel(message)) {
             logger.warn(
                 `${getDebugLogHeader(
                     message
-                )} | Invalid force-skip. !gameSession: ${!gameSession}. !gameSession.round: ${
-                    gameSession && !gameSession.round
-                }. !areUserAndBotInSameVoiceChannel: ${!areUserAndBotInSameVoiceChannel(
-                    message
-                )}`
+                )} | Invalid forceskip. User and bot are not in the same voice channel.`
             );
             return;
         }
 
-        if (gameSession.round.skipAchieved) {
-            // song already being skipped
+        const guildPreference = await getGuildPreference(message.guildID);
+        const session = Session.getSession(message.guildID);
+        if (
+            !session ||
+            !session.round ||
+            session.round.skipAchieved ||
+            session.round.finished ||
+            !areUserAndBotInSameVoiceChannel(message)
+        ) {
             return;
         }
 
-        if (message.author.id !== gameSession.owner.id) {
+        if (message.author.id !== session.owner.id) {
             await sendErrorMessage(MessageContext.fromMessage(message), {
                 title: state.localizer.translate(
                     message.guildID,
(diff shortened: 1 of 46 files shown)


src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+    aliases = ["radio", "listen"];
+
+    help = (guildID: string): Help => ({
+        name: "music",
+        description: state.localizer.translate(
+            guildID,
+            "command.music.help.description"
+        ),
+        usage: "music",
+        // priority: 1050,
```
taahamahdi, line 91: TODO

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+        arguments: [],
+    };
+
+    aliases = ["radio", "listen"];
+
+    help = (guildID: string): Help => ({
+        name: "music",
+        description: state.localizer.translate(
+            guildID,
+            "command.music.help.description"
```
taahamahdi, line 88: todo

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+        ),
+        usage: "music",
+        // priority: 1050,
+        examples: [
+            {
+                example: "`,music`",
+                explanation: state.localizer.translate(
+                    guildID,
+                    "command.music.help.example"
```
taahamahdi, line 97: todo

src/commands/game_commands/forceskip.ts
```diff
@@ -35,32 +36,29 @@ export default class ForceSkipCommand implements BaseCommand {
...
             );
             return;
         }
 
-        if (gameSession.round.skipAchieved) {
-            // song already being skipped
+        const guildPreference = await getGuildPreference(message.guildID);
+        const session = Session.getSession(message.guildID);
+        if (
+            !session ||
```
Brainicism, line 52: `inSessionCommandPrecheck` already checks for !session

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @param participants - The list of participants
+ */
+export async function sendBeginGameMessage(
+    textChannelName: string,
+    voiceChannelName: string,
+    message: GuildTextableMessage
+): Promise<void> {
+    const startTitle = state.localizer.translate(
+        message.guildID,
+        "command.play.gameStarting",
```
Brainicism, line 41: does gameSTarting make sense?

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+            voiceChannelName,
+        }
+    );
+
+    const gameInfoMessage: GameInfoMessage = chooseWeightedRandom(
+        await dbContext.kmq("game_messages")
+    );
+
+    const fields: Eris.EmbedField[] = [];
+    if (gameInfoMessage) {
```
Brainicism, line 53: Will this ever be null?

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Sends the beginning of game session message
+ * @param textChannelName - The name of the text channel to send the message to
+ * @param voiceChannelName - The name of the voice channel to join
+ * @param message - The original message that triggered the command
+ * @param participants - The list of participants
+ */
+export async function sendBeginGameMessage(
+    textChannelName: string,
```
Brainicism, line 35: does beginGame make sense here?

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+                description: "command.music.failure.existingGameSession.title",
+            });
+            return;
+        }
+
+        const guildPreference = await getGuildPreference(message.guildID);
+        const textChannel = channel;
+        const gameOwner = KmqMember.fromUser(message.author);
+        const voiceChannel = getUserVoiceChannel(messageContext);
+        const timeUntilRestart = await getTimeUntilRestart();
```
Brainicism, line 119: Is this copied from play.ts? If so, can we move it somewhere common

src/commands/game_commands/music.ts
```diff
@@ -0,0 +1,182 @@
...
+                    message.guildID,
+                    "command.play.failure.botRestarting.description",
+                    { timeUntilRestart: `\`${timeUntilRestart}\`` }
+                ),
+            });
+
+            logger.warn(
+                `${getDebugLogHeader(
+                    message
+                )} | Attempted to start game before restart.`
```
Brainicism, line 136: session, not game

src/commands/game_commands/play.ts
```diff
@@ -210,24 +209,24 @@ export default class PlayCommand implements BaseCommand {
...
         parsedMessage,
         channel,
     }: CommandArgs): Promise<void> => {
-        const guildPreference = await getGuildPreference(message.guildID);
-        const voiceChannel = getUserVoiceChannel(
-            MessageContext.fromMessage(message)
-        );
+        const messageContext = MessageContext.fromMessage(message);
+        const guildID = message.guildID;
+        const existingGameSession = Session.getSession(guildID) as GameSession;
```
Brainicism, line 217: The reason we use `gameSessions[message.guildID]` everywhere instead of saving it to a variable is to account for race conditions where the session is deleted before the line runs. Can you make sure it's not an issue here before making this change?

Brainicism (changes requested):

src/commands/game_commands/music.ts
taahamahdi: Similar behavior copied over from play.ts. Technically the table can be unpopulated, but I can remove it from both files

src/commands/game_commands/music.ts
taahamahdi: It is, I'll move it to a precheck

src/commands/game_commands/play.ts
taahamahdi: Forgot about that, I can revert these changes

src/commands/game_options/special.ts
```diff
@@ -64,6 +64,7 @@ export default class SpecialCommand implements BaseCommand {
                 "This is an unreleased game option, and can only be used on the official KMQ server",
```
taahamahdi, line 64: Need to figure out a way to precheck if either premium or debug server

src/structures/session.ts
```diff
@@ -351,7 +365,11 @@ export default abstract class Session {
         messageContext: MessageContext,
```
taahamahdi, line 365: move this fn to game session since its not used in music

Brainicism (approved):

PR #1266, opened 2022-04-09
Deprecate random shuffle with unique

Brainicism:


src/migrations/20220409183111_deprecate_old_shuffle.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.raw(
```
taahamahdi, line 6: Alternatively, you can delete all shuffletype `game_options`

taahamahdi (approved):

src/migrations/20220409183111_deprecate_old_shuffle.js
Brainicism: Would rather not delete since the existence of the option tells us a little bit about how many servers have played since the option was introduced

PR #1281, opened 2022-04-17
Cyclic dependency cleanup: Extract exported interfaces

Brainicism:
#1280 
Reduces warnings from 410 to 391

### DIFF
--- a/src/interfaces/state.ts
+++ b/src/interfaces/state.ts
@@ -8,7 +8,7 @@ import GameSession from "../structures/game_session";
 import MusicSession from "../structures/music_session";
 import { Campaign } from "patreon-discord";
 
-export interface State {
+export default interface State {
     gameSessions: { [guildID: string]: GameSession };
     musicSessions: { [guildID: string]: MusicSession };
     client: KmqClient;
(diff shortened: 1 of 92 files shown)


src/interfaces/state.ts
```diff
@@ -0,0 +1,26 @@
...
+import LocalizationManager, {
+    LocaleType,
+} from "../helpers/localization_manager";
+import KmqClient from "../kmq_client";
+import RateLimiter from "../rate_limiter";
+import GameSession from "../structures/game_session";
+import MusicSession from "../structures/music_session";
+import { Campaign } from "patreon-discord";
+
+export interface State {
```
taahamahdi, line 11: Can probably be default

taahamahdi (dismissed):

src/kmq_worker.ts
```diff
@@ -21,6 +21,7 @@ import ReloadCommand from "./commands/admin/reload";
 import EvalCommand from "./commands/admin/eval";
 import LocalizationManager from "./helpers/localization_manager";
 import Session from "./structures/session";
+import State from "./interfaces/state";
```
taahamahdi, line 24: I think this can be imported directly since it's a default export

taahamahdi (changes requested):

taahamahdi (approved):

PR #1291, opened 2022-04-24
Cyclic dependency cleanup: Extract exported consts

Brainicism:
#1280 
Reduces warnings from 41 to 24


src/constants.ts
```diff
@@ -93,3 +96,112 @@ export const CUM_EXP_TABLE = EXP_TABLE.map(
             (sum += value)
     )(0)
 );
+
+export const VOTE_BONUS_DURATION = 1;
```
taahamahdi, line 100: Should we move related values into namespaces here?

Brainicism: Eventually, this can come later though

taahamahdi (approved):

PR #1293, opened 2022-04-24
Cyclic dependency cleanup: Remove dependence on Session in discord_utils

Brainicism:
#1280 
Reduces warnings from 6 to 5


src/helpers/discord_utils.ts
```diff
@@ -898,13 +899,13 @@ export function getFormattedLimit(
  *  @returns an embed of current game options
  */
```
taahamahdi, line 900: isMusicSession can be removed here

taahamahdi (approved):

PR #1323, opened 2022-05-04
Add song name search functionality to lookup

Brainicism:
If user provides a valid URL, we assume that they are providing a youtube link and look up by youtube video ID
Otherwise, we assume they're providing a song name, and we search using song name against `available_songs.` If there are multiple results, we list the matched songs, including pagination if needed. Matches are sorted by song name length, then by views.

image

If there is only one result, we return a response similar to searching by video ID

### DIFF
--- a/src/commands/game_commands/lookup.ts
+++ b/src/commands/game_commands/lookup.ts
@@ -47,10 +47,10 @@ const getDaisukiLink = (id: string, isMV: boolean): string => {
 async function lookupByYoutubeID(
     message: GuildTextableMessage,
     videoID: string,
-    guildID: string,
     locale: LocaleType
 ): Promise<void> {
     const messageContext = MessageContext.fromMessage(message);
+    const guildID = message.guildID;
     const kmqSongEntry: QueriedSong = await dbContext
         .kmq("available_songs")
         .select(SongSelector.getQueriedSongFields())
@@ -251,7 +251,7 @@ async function lookupByYoutubeID(
     }
 
     sendInfoMessage(messageContext, {
-        title: `${songName} - ${artistName}`,
+        title: `**${songName}** - ${artistName}`,
         url: `https://youtu.be/${videoID}`,
         description,
         thumbnailUrl: `https://img.youtube.com/vi/${videoID}/hqdefault.jpg`,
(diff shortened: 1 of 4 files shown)


src/commands/game_commands/lookup.ts
```diff
@@ -78,263 +362,51 @@ export default class LookupCommand implements BaseCommand {
...
-
-            logger.info(
-                `${getDebugLogHeader(
-                    message
-                )} | Non-KMQ song lookup. videoID = ${videoID}.`
-            );
-        }
+        if (isValidURL(arg)) {
+            let videoID: string = null;
+            if (arg.startsWith("<") && arg.endsWith(">")) {
```
taahamahdi, line 371: Is it still considered a valid url if it starts with < and ends with > ? Might need to move this before the if statement

src/commands/game_commands/lookup.ts
```diff
@@ -39,11 +44,290 @@ const getDaisukiLink = (id: string, isMV: boolean): string => {
     return `https://kpop.daisuki.com.br/audio_videos.html?playid=${id}`;
 };
 
+async function lookupByYoutubeID(
+    message: GuildTextableMessage,
+    videoID: string,
+    guildID: string,
```
taahamahdi, line 50: I think the message argument might contain the guild ID

src/commands/game_commands/lookup.ts
```diff
@@ -39,11 +44,290 @@ const getDaisukiLink = (id: string, isMV: boolean): string => {
...
+): Promise<void> {
+    const messageContext = MessageContext.fromMessage(message);
+    const kmqSongEntries: QueriedSong[] = await dbContext
+        .kmq("available_songs")
+        .select(SongSelector.getQueriedSongFields())
+        .whereRaw(`song_name_en LIKE "%${songName}%"`)
+        .orWhereRaw(`song_name_ko LIKE "%${songName}%"`)
+        .orderByRaw("CHAR_LENGTH(song_name_en) ASC")
+        .orderBy("views", "DESC")
+        .limit(100);
```
taahamahdi, line 280: Will the number of results impact performance?

src/commands/game_commands/lookup.ts
```diff
@@ -39,11 +44,290 @@ const getDaisukiLink = (id: string, isMV: boolean): string => {
...
+
+    if (kmqSongEntries.length === 0) {
+        await sendInfoMessage(messageContext, {
+            title: "Lookup Results",
```
taahamahdi, line 284: Could be localized

src/commands/game_commands/lookup.ts
```diff
@@ -39,11 +44,290 @@ const getDaisukiLink = (id: string, isMV: boolean): string => {
...
+                value: LocalizationManager.localizer.translate(
+                    guildID,
+                    includedInOptions ? "misc.yes" : "misc.no"
+                ),
+            }
+        );
+    }
+
+    sendInfoMessage(messageContext, {
+        title: `${songName} - ${artistName}`,
```
taahamahdi, line 254: Nit: The song name is not bold here but is bold in the results, which works better?

taahamahdi (changes requested):

src/commands/game_commands/lookup.ts
Brainicism: At 100, no

edwardhdlu:
pog

taahamahdi (approved):

sabihismail:
pog

PR #1343, opened 2022-05-15
Store game messages + translations in database

Brainicism:


src/scripts/game-message-converter.ts
```diff
@@ -0,0 +1,44 @@
...
+const enFile = JSON.parse(
+    fs.readFileSync(path.resolve(__dirname, "../../i18n/en.json")).toString()
+);
+
+const koFile = JSON.parse(
+    fs.readFileSync(path.resolve(__dirname, "../../i18n/ko.json")).toString()
+);
+
+(async () => {
+    if (require.main === module) {
```
taahamahdi, line 18: Early return instead

src/migrations/20220514235230_game-messages-varchar-to-text.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+    return knex.schema.table("game_messages", function (table) {
+        table.dropColumn("id");
```
taahamahdi, line 26: When is this column added

taahamahdi (approved):

src/migrations/20220514235230_game-messages-varchar-to-text.js
Brainicism: `table.increments();` creates a auto-increment primary key called 'id'

src/scripts/game-message-converter.ts
Brainicism: This is how we do it everywhere

PR #1353, opened 2022-05-20
Add soft restarts for minimal downtime

Brainicism:


src/scripts/announce-restart.ts
```diff
@@ -54,6 +71,8 @@ process.on("SIGINT", async () => {
     const restartDate = new Date();
     restartDate.setMinutes(restartDate.getMinutes() + restartMinutes);
 
+    console.log(options);
```
taahamahdi, line 74:
```suggestion
```

.github/workflows/deploy-prod.yml
```diff
@@ -1,45 +1,54 @@
...
-        uses: actions/checkout@v2
+    snap:
+        runs-on: ubuntu-latest
+        timeout-minutes: 10
+        steps:
+            - name: Checkout KMQ_Discord
+              uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+                  token: ${{ secrets.PAT }}
```
taahamahdi, line 30: I think I missed where secrets.PAT is declared

taahamahdi (changes requested):

src/scripts/announce-restart.ts
Brainicism: It's a nice to have

.github/workflows/deploy-prod.yml
Brainicism: This wasn't added in this commit, it's in the repo's secret key vault

taahamahdi (approved):

PR #1368, opened 2022-06-11
Use nicknames for players, various player cleanup

### LINKED ISSUE #1365: Show user tag instead of username in scoreboard if user is not active on scoreboard

taahamahdi:
Closes #1365.

### DIFF
--- a/src/structures/player.ts
+++ b/src/structures/player.ts
@@ -4,15 +4,14 @@ import ExpBonusModifier from "../enums/exp_bonus_modifier";
 import State from "../state";
 
 export default class Player {
-    /** The Discord username of the player sans discriminator,
-     * i.e. "Player" when the player's user tag is "Player#1234"
-     */
-    public readonly name: string;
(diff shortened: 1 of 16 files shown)


src/structures/player.ts
```diff
@@ -102,6 +102,17 @@ export default class Player {
         return name;
     }
 
+    /**
+     * @param guildID - The guild ID
+     * @returns the player's nickname in the given guild
+     */
+    getName(): string {
+        return (
+            State.client.guilds.get(this.guildID).members.get(this.id).nick ??
```
Brainicism, line 111: Why retrieve them from state here, instead of passing them in? Should try to avoid global accesses in our plain classes

taahamahdi: If a user changes their nickname mid-game, it can become confusing who is who. Some games run for hours, which can make this worse

Brainicism (approved):

PR #1384, opened 2022-07-01
fixed the quote issue in lookup.ts

kambojk:
### What will this change affect?
You see, before this change, we had the unfortunate realization that this menu was causing some players discomfort. This is because this specific message was an outlier, which bothered quite a few users. This change will increase player morale, and allow players to focus on the real important part of KMQ, the game.

### Testing?
I'm sorry I didn't test it.


taahamahdi (approved):

sabihismail:
bazed

PR #1385, opened 2022-07-03
Cache username on Player creation

taahamahdi:
If a player is uncached (e.g. they leave the server), then accessing their username from the bot's cache will fail. Instead, we store the username once at the start of each game

### DIFF
--- a/src/structures/player.ts
+++ b/src/structures/player.ts
@@ -7,6 +7,9 @@ export default class Player {
     /** The Discord user ID of the player */
     public readonly id: string;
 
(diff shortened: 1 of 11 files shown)


src/structures/player.ts
```diff
@@ -106,7 +112,7 @@ export default class Player {
     getName(): string {
         return (
             State.client.guilds.get(this.guildID).members.get(this.id).nick ??
```
Brainicism, line 114: Don't we have the same issue here?

PR #1431, opened 2022-07-15
Allow reload/delete of single command

taahamahdi:


src/commands/admin/appcommands.ts
```diff
@@ -120,17 +137,24 @@ export default class AppCommandsCommand implements BaseCommand {
 
             sendInfoMessage(MessageContext.fromMessage(message), {
                 title: "Application Commands Reloaded",
-                description: "Yay.",
+                description: commandsReloaded.join(", "),
             });
         } else {
-            const commands = await State.client.getCommands();
+            const commandsToModify = isSingleCommand
```
Brainicism, line 143: This is common between reload and delete

Brainicism (changes requested):

src/commands/admin/appcommands.ts
taahamahdi: Object.keys vs Object.values

Brainicism (approved):

Brainicism (approved):

PR #1433, opened 2022-07-16
Update review link (top.gg anchor no longer works)

taahamahdi:


src/constants.ts
```diff
@@ -104,7 +104,8 @@ export const CUM_EXP_TABLE = EXP_TABLE.map(
 
 export const VOTE_BONUS_DURATION = 1;
 export const VOTE_LINK = "https://top.gg/bot/508759831755096074/vote";
-export const REVIEW_LINK = "https://top.gg/bot/508759831755096074#reviews";
+export const REVIEW_LINK =
```
Brainicism, line 107: Does this work across all major browsers? Test on non-chromium

Brainicism (changes requested):

src/constants.ts
taahamahdi: This will only work in Chromium-based browsers. Top.gg doesn't support an official link to reviews anymore, so this might be our best bet

Brainicism (approved):

PR #1438, opened 2022-07-24
Re-use send(Info/Error)Message for interaction responses

Brainicism:


src/helpers/discord_utils.ts
```diff
@@ -376,13 +376,29 @@ async function sendMessageExceptionHandler(
...
+    authorID?: string,
+    interaction?: Eris.ComponentInteraction | Eris.CommandInteraction
 ): Promise<Eris.Message> {
+    if (interaction) {
+        if (!withinInteractionInterval(interaction)) {
+            return null;
+        }
+
+        try {
+            await interaction.createMessage(messageContent);
```
taahamahdi, line 394: should we return the message?

taahamahdi (changes requested):

src/helpers/discord_utils.ts
Brainicism: it returns Promise<void> and we don't ever use the return value

taahamahdi: This function returns `Promise<Eris.Message>`

Brainicism: https://abal.moe/Eris/docs/0.17.1/CommandInteraction#method-createMessage
image
image

taahamahdi: Whoops, my bad

taahamahdi (approved):

PR #1441, opened 2022-07-25
Slash command for ,goal

Brainicism:


hoehoetvhss (dismissed):

src/commands/game_options/goal.ts
```diff
@@ -66,88 +69,166 @@ export default class GoalCommand implements BaseCommand {
...
+            type: Eris.Constants.ApplicationCommandTypes.CHAT_INPUT,
+            options: [
+                {
+                    name: "goal",
+                    description: LocalizationManager.localizer.translate(
+                        LocaleType.EN,
+                        "command.goal.interaction.description"
+                    ),
+                    type: Eris.Constants.ApplicationCommandOptionTypes.INTEGER,
+                    required: true,
```
taahamahdi, line 88: Should we make it disable the goal if no args are passed? Which would mean `required: false`

taahamahdi (changes requested):

src/commands/game_options/goal.ts
```diff
@@ -66,88 +69,166 @@ export default class GoalCommand implements BaseCommand {
...
                 return;
             }
         }
 
-        await guildPreference.setGoal(userGoal);
+        const reset = userGoal === null;
+
+        if (reset) {
+            await guildPreference.reset(GameOption.GOAL);
+            await sendOptionsMessage(
```
taahamahdi, line 188: I think this can be removed, otherwise it will send the options message twice on reset

taahamahdi (changes requested):

taahamahdi (approved):

PR #1445, opened 2022-07-25
Slash command for ,listen

Brainicism:


src/commands/game_commands/listen.ts
```diff
@@ -133,9 +146,29 @@ export default class ListenCommand implements BaseCommand {
...
+    call = async ({ message }: CommandArgs): Promise<void> => {
+        await ListenCommand.startListening(MessageContext.fromMessage(message));
+    };
+
+    static startListening = async (
+        messageContext: MessageContext,
+        interaction?: Eris.CommandInteraction
+    ): Promise<void> => {
+        if (interaction) {
+            await tryCreateInteractionSuccessAcknowledgement(
```
taahamahdi, line 158: We should pass the interaction as a new argument to `sendBeginListeningSessionMessage` and use that as the interaction acknowledgement

taahamahdi (changes requested):

taahamahdi (approved):

hoehoetvhss (approved):

PR #1451, opened 2022-08-06
Daisuki seed interval defined in cronjob

taahamahdi:
https://stackoverflow.com/a/9393636

`DAISUKI_SEED_CRON_JOB="15 3,15 * * *"` set in prod


src/kmq.ts
```diff
@@ -90,8 +90,19 @@ function registerGlobalIntervals(fleet: Fleet): void {
...
+            await dbContext.kmq("system_stats").insert({
+                stat_name: "request_latency",
+                stat_value: fleet.eris.requestHandler.latencyRef.latency,
+                date: new Date(),
+            });
+        }
+    });
+
+    // as defined in DAISUKI_SEED_CRON_JOB
+    schedule.scheduleJob(process.env.DAISUKI_SEED_CRON_JOB, async () => {
```
Brainicism, line 105: Add default if not specified

taahamahdi: It's a required env var

Brainicism: yes

Brainicism (approved):

PR #1454, opened 2022-08-06
Slash command for ,duration

Brainicism:


src/commands/game_options/duration.ts
```diff
@@ -94,80 +107,226 @@ export default class DurationCommand implements BaseCommand {
...
+                                LocalizationManager.localizer.translate(
+                                    LocaleType.EN,
+                                    "command.duration.help.example.increment"
+                                ),
+                            required: true,
+                            min_value: DURATION_DELTA_MIN,
+                            max_value: DURATION_DELTA_MAX,
+                            type: Eris.Constants.ApplicationCommandOptionTypes
+                                .INTEGER,
+                        },
```
taahamahdi, line 164: Will this need a cast to `any`?

src/commands/game_options/duration.ts
```diff
@@ -94,80 +107,226 @@ export default class DurationCommand implements BaseCommand {
...
+                                LocalizationManager.localizer.translate(
+                                    LocaleType.EN,
+                                    "command.duration.help.example.decrement"
+                                ),
+                            required: true,
+                            min_value: DURATION_DELTA_MIN,
+                            max_value: DURATION_DELTA_MAX,
+                            type: Eris.Constants.ApplicationCommandOptionTypes
+                                .INTEGER,
+                        },
```
taahamahdi, line 188: Same here

src/commands/game_options/duration.ts
Brainicism: Yes until we update Eris to fix their typings

taahamahdi (approved):

PR #1455, opened 2022-08-07
Slash command for ,preset

Brainicism:


hoehoetvhss (approved):

src/commands/game_commands/preset.ts
```diff
@@ -118,132 +215,202 @@ export default class PresetCommand implements BaseCommand {
...
+                    ),
+                    type: Eris.Constants.ApplicationCommandOptionTypes
+                        .SUB_COMMAND,
+                    options: [
+                        {
+                            name: "preset_name",
+                            description:
+                                LocalizationManager.localizer.translate(
+                                    LocaleType.EN,
+                                    "command.preset.help.example.save"
```
taahamahdi, line 250: May need a new translation string for the slash command options based on the context
save : the name to save the preset as
load/delete: the name of the preset to load/delete
replace: the preset to replace
import: the unique identifier associated with the preset

taahamahdi (changes requested):

hoehoetvhss (approved):

PR #1459, opened 2022-08-11
Allow Spotify playlist for custom song list

taahamahdi:


hoehoetvhss (dismissed):

src/helpers/discord_utils.ts
```diff
@@ -898,6 +944,13 @@ export async function generateOptionsMessage(
         .filter((option) => optionStrings[option as GameOption])
         .filter((option) => !PriorityGameOption.includes(option as GameOption));
 
+    // Move answer out from the priority options
+    if (isSpotify) {
```
Brainicism, line 948: What is this doing?

src/structures/song_selector.ts
```diff
@@ -191,35 +198,46 @@ export default class SongSelector {
 
     async reloadSongs(
         guildPreference: GuildPreference,
-        isPremium: boolean
+        isPremium: boolean,
+        songList?: MatchedPlaylist
```
Brainicism, line 202: Doesn't really make sense for a function called `reloadSongs` to take in a list of songs to passthrough?

src/commands/game_options/spotify.ts
```diff
@@ -0,0 +1,290 @@
...
+        if (parsedMessage.components.length === 0) {
+            playlistURL = null;
+            await SpotifyCommand.updateOption(
+                MessageContext.fromMessage(message),
+                playlistURL
+            );
+        } else if (isValidURL(parsedMessage.components[0])) {
+            await SpotifyCommand.updateOption(
+                MessageContext.fromMessage(message),
+                parsedMessage.components[0]
```
Brainicism, line 102: Can use playlistURL and make one call for both of these two cases

src/commands/game_options/spotify.ts
```diff
@@ -0,0 +1,290 @@
...
+        const gameSession = State.gameSessions[guildID];
+        const reset = playlistURL === null;
+        if (reset) {
+            await guildPreference.reset(GameOption.SPOTIFY_PLAYLIST_METADATA);
+            logger.info(
+                `${getDebugLogHeader(messageContext)} | Spotify playlist reset.`
+            );
+        } else if (
+            isValidURL(playlistURL) &&
+            playlistURL.toLocaleLowerCase().startsWith(SPOTIFY_BASE_URL) &&
```
Brainicism, line 137: Better to do a regex?

src/helpers/spotify_manager.ts
```diff
@@ -0,0 +1,237 @@
...
+     */
+    getMatchedSpotifySongs = async (
+        playlistID: string,
+        isPremium: boolean
+    ): Promise<MatchedPlaylist> => {
+        if (
+            !process.env.SPOTIFY_CLIENT_ID ||
+            !process.env.SPOTIFY_CLIENT_SECRET
+        ) {
+            return {
```
Brainicism, line 58: Log warning

src/helpers/spotify_manager.ts
```diff
@@ -0,0 +1,237 @@
...
+                const response = (
+                    await Axios.get(requestURL, {
+                        headers: {
+                            Authorization: `Bearer ${this.accessToken}`,
+                            "Content-Type": "application/x-www-form-urlencoded",
+                        },
+                    })
+                ).data;
+
+                for (const item of response.items) {
```
Brainicism, line 85: map?

src/helpers/spotify_manager.ts
```diff
@@ -0,0 +1,237 @@
...
+                logger.error(`Failed fetching Spotify playlist. err = ${err}`);
+
+                if (err.response) {
+                    logger.info(err.response.data);
+                    logger.info(err.response.status);
+                }
+
+                break;
+            }
+        } while (requestURL);
```
Brainicism, line 105: What are spotify rate limits like?

src/helpers/spotify_manager.ts
```diff
@@ -0,0 +1,237 @@
...
+                        : process.env.AUDIO_SONGS_PER_ARTIST
+                )
+                .first()) as QueriedSong;
+
+            if (result) {
+                matchedSongs.push(result);
+            }
+        }
+
+        const spotifyMetadata = await this.getPlaylistMetadata(playlistID);
```
Brainicism, line 160: This must return undefined if there was an error

Brainicism (changes requested):

src/structures/song_selector.ts
taahamahdi: We'll need to retrieve the playlist metadata from the `,spotify` command, so I pass it along here to prevent hitting the endpoint twice

src/helpers/discord_utils.ts
taahamahdi: This clears the first options field and moves the ,spotify/,answer fields to the start. I'll update the comment.

src/helpers/spotify_manager.ts
taahamahdi: They aren't well defined, but the response will contain a 429 and a `Retry-After` header. I'll check for that scenario and use our `retryJob` function

Brainicism (approved):

PR #1486, opened 2022-09-17
Add text command deprecation feature switches

Brainicism:


src/events/client/messageCreate.ts
```diff
@@ -94,6 +97,43 @@ export default async function messageCreateHandler(
...
+                    message.guildID,
+                    "misc.failure.slashCommandReminder.title"
+                ),
+                description: LocalizationManager.localizer.translate(
+                    message.guildID,
+                    "misc.failure.slashCommandReminder.description"
+                ),
+            });
+
+            await delay(1500);
```
taahamahdi, line 134: This delay should be removed, it will annoy the user

taahamahdi (changes requested):

src/events/client/messageCreate.ts
Brainicism: The message gets flooded right away if there's no delay, and its 1 in 10 messages for a very important deprecation

taahamahdi (dismissed):

PR #1518, opened 2022-12-06
Support slash command localization

taahamahdi:


taahamahdi:
@hoehoetvhss When you get a chance, can you take out the relevant `separated by commas` for the changes in this commit: https://github.com/Brainicism/KMQ_Discord/pull/1518/commits/a146520a62ba3f7c01b7f904e3bd9caefd04e1ce

taahamahdi:
Looks good, thanks! I've removed the whitespaces from the `name` fields because Discord expects these to be single words. Can you verify the command names still make sense in this commit? https://github.com/Brainicism/KMQ_Discord/pull/1518/commits/48fcec3c68ae577713de3b13b06462751de59e3b

hoehoetvhss:
The current translation is very different from the past. That's why everyone is making corrections, but it takes time.

hoehoetvhss:
> The current translation is very different from the past. That's why everyone is making corrections, but it takes time.

But it doesn't interfere with the use.

src/commands/admin/appcommands.ts
```diff
@@ -95,10 +100,47 @@ export default class AppCommandsCommand implements BaseCommand {
...
-                    const commands = command.slashCommands();
+                    const commands =
+                        command.slashCommands() as Array<Eris.ChatInputApplicationCommandStructure>;
+
                     for (const cmd of commands) {
+                        cmd.name =
+                            cmd.name ??
+                            i18n.translate(
+                                LocaleType.EN,
+                                `command.${commandName}.help.name`
```
Brainicism, line 115: Can we throw an exception if the 'name' translation key is missing, but we're expecting it to exist? Or a test

Brainicism (changes requested):

hoehoetvhss:
Any parts that I do not understand or are difficult to translate will be left in English for later translation.

Brainicism (approved):

PR #1560, opened 2023-01-31
Strict mode: strictNullChecks part 3 (structures/helpers)

Brainicism:


src/command_prechecks.ts
```diff
@@ -47,7 +47,7 @@ export default class CommandPrechecks {
             messageContext.guildID
         );
 
-        if (session.isListeningSession()) {
+        if (session.isListeningSession() && interaction) {
```
taahamahdi, line 50: What about an in session check triggered by a non-interaction command, with the bot and user in different channels? We'd skip entering here and would be casting the ListeningSession as a GameSession

src/events/client/interactionCreate.ts
```diff
@@ -104,6 +104,7 @@ export default async function interactionCreateHandler(
                 logger.error(
                     `No handler found for CHAT_INPUT CommandInteraction: ${interaction.data.name}`
                 );
+                return;
```
taahamahdi, line 107: Are the early returns needed for `No handler found` cases? We already return if `interactionPromise === null`

src/helpers/validate.ts
```diff
@@ -94,7 +94,7 @@ export default (
 
                 // parse as integer for now, might cause problems later?
                 const intArg = parseInt(arg, 10);
-                if ("minValue" in validation && intArg < validation.minValue) {
+                if (validation.minValue && intArg < validation.minValue) {
```
taahamahdi, line 97: If `minValue` is zero this check never continues

src/helpers/validate.ts
```diff
@@ -111,7 +111,7 @@ export default (
                     return false;
                 }
 
-                if ("maxValue" in validation && intArg > validation.maxValue) {
+                if (validation.maxValue && intArg > validation.maxValue) {
```
taahamahdi, line 114: If `maxValue` is zero this check never continues (less likely but still possible)

src/kmq.ts
```diff
@@ -180,15 +201,18 @@ async function startWebServer(fleet: Fleet): Promise<void> {
             return;
         }
 
-        const isSoftRestart = request.body["soft"] ? 1 : 0;
+        const isSoftRestart = (request.body as any)["soft"] ? 1 : 0;
```
taahamahdi, line 204: Nit: Can cast to an object with `soft` property

src/kmq.ts
```diff
@@ -180,15 +201,18 @@ async function startWebServer(fleet: Fleet): Promise<void> {
...
 
         if (isSoftRestart) {
             logger.info("Soft restart initiated");
             await exec("npx tsc");
             fleet.restartAllClusters(false);
         }
 
-        const restartMinutes = request.body["restartMinutes"];
+        const restartMinutes = (request.body as any)[
```
taahamahdi, line 212: Nit: Can cast to an object with `restartMinutes` property

src/kmq.ts
```diff
@@ -216,7 +240,7 @@ async function startWebServer(fleet: Fleet): Promise<void> {
             return;
         }
 
-        const userID = request.body["user"];
+        const userID = (request.body as any)["user"] as string;
```
taahamahdi, line 243: Nit: Can cast to an object with `user` property

src/seed/bootstrap.ts
```diff
@@ -69,8 +69,8 @@ async function songThresholdReached(db: DatabaseContext): Promise<boolean> {
     if (!availableSongsTableExists) return false;
 
     return (
-        (await db.kmq("available_songs").count("* as count").first()).count >=
-        SONG_DOWNLOAD_THRESHOLD
+        ((await db.kmq("available_songs").count("* as count").first()) as any)
```
taahamahdi, line 72: Nit: Can cast to an object with `count` property

src/seed/seed_db.ts
```diff
@@ -63,12 +63,12 @@ export async function tableExists(
...
     return (
         (
-            await db
+            (await db
                 .agnostic("information_schema.tables")
                 .where("table_schema", "=", databaseName)
                 .where("table_name", "=", tableName)
                 .count("* as count")
-                .first()
+                .first()) as any
```
taahamahdi, line 71: Nit: Can cast to an object with `count` property

src/seed/seed_db.ts
```diff
@@ -247,32 +247,32 @@ async function validateSqlDump(
 
         logger.info("Validating MV song count");
         const mvSongCount = (
-            await db
+            (await db
                 .kpopVideosValidation("app_kpop")
                 .count("* as count")
                 .where("is_audio", "=", "n")
-                .first()
+                .first()) as any
```
taahamahdi, line 254: Nit: Can cast to an object with `count` property

src/seed/seed_db.ts
```diff
@@ -247,32 +247,32 @@ async function validateSqlDump(
...
 
         logger.info("Validating audio-only song count");
         const audioSongCount = (
-            await db
+            (await db
                 .kpopVideosValidation("app_kpop")
                 .count("* as count")
                 .where("is_audio", "=", "y")
-                .first()
+                .first()) as any
```
taahamahdi, line 265: Nit: Can cast to an object with `count` property

src/seed/seed_db.ts
```diff
@@ -247,32 +247,32 @@ async function validateSqlDump(
...
         logger.info(`Found ${audioSongCount} audio-only videos`);
 
         logger.info("Validating group count");
         const artistCount = (
-            await db
+            (await db
                 .kpopVideosValidation("app_kpop_group")
                 .count("* as count")
-                .first()
+                .first()) as any
```
taahamahdi, line 275: Nit: Can cast to an object with `count` property

src/structures/listening_session.ts
```diff
@@ -156,11 +154,11 @@ export default class ListeningSession extends Session {
         }
 
         const round = this.round;
-        const guildID = interaction.guildID;
+        const guildID = interaction.guildID as string;
         if (interaction.data.custom_id === "bookmark") {
             this.handleBookmarkInteraction(interaction);
```
taahamahdi, line 159: Do we need to await this?

taahamahdi (changes requested):

src/command_prechecks.ts
Brainicism: This conditional is only for interaction response via `tryCreateInteractionErrorAcknowledgement`. If we want to make it the same behavior for non-interactions then we need another PR for it.

src/structures/listening_session.ts
Brainicism: Don't know but not within the scope of this PR

taahamahdi: This PR makes `handleBookmarkInteraction` async so we might need to await where appropriate

taahamahdi (approved):

PR #1575, opened 2023-02-11
Improve spotify song matching

Brainicism:
- Better matching for song names with brackets. e.g "Egoist (Olivia Hye)". Attempts to match each component separately, as well as all together
- Match artists in all lowercase AND punctuation removed
- Match song names with all non-alphanumeric characters removed
- Match against id_parent_artist

### DIFF
--- a/src/helpers/game_utils.ts
+++ b/src/helpers/game_utils.ts
@@ -609,7 +609,5 @@ export function isPowerHour(): boolean {
  * @returns the normalized artist name for state entry
  */
 export function normalizeArtistNameEntry(artistName: string): string {
-    return normalizePunctuationInName(
-        artistName.toLowerCase().replaceAll(" ", "")
-    );
+    return normalizePunctuationInName(artistName.toLowerCase());
 }
--- a/src/helpers/game_utils.ts
+++ b/src/helpers/game_utils.ts
@@ -3,7 +3,10 @@ import {
     PATREON_SUPPORTER_BADGE_ID,
     SHADOW_BANNED_ARTIST_IDS,
 } from "../constants";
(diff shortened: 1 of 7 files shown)


src/helpers/game_utils.ts
```diff
@@ -600,3 +603,13 @@ export function isPowerHour(): boolean {
...
     );
 }
+
+/**
+ * @param artistName - the artist name
+ * @returns the normalized artist name for state entry
+ */
+export function normalizeArtistNameEntry(artistName: string): string {
+    return normalizePunctuationInName(
+        artistName.toLowerCase().replaceAll(" ", "")
```
taahamahdi, line 613: `normalizePunctuationInName` already removes spaces since it's a REMOVED_CHARACTER:

```ts
const REMOVED_CHARACTERS = /[\|’\ '?!.\-,:;★*´\(\)\+\u200B…]/g;
```

src/helpers/spotify_manager.ts
```diff
@@ -198,26 +204,47 @@ export default class SpotifyManager {
                 }
             }
 
-            const result = (await dbContext
+            // handle songs with brackets in name, consider all components separately
+            const songNameBracketComponents = song.name.split("(");
+            const songNames = [songNameBracketComponents[0], song.name];
```
taahamahdi, line 209: Does the song name get added to this twice if there isn't a `(` in the name? If so, you could move pushing `songNameBracketComponents[0]` on the `songNames` array into the if statement, so we end up doing less `orWhere`

taahamahdi (approved):

taahamahdi (approved):

PR #1599, opened 2023-03-05
Substitute songs with better audio counterparts

Brainicism:


x/main.js
```diff
@@ -0,0 +1,54 @@
+const ytdl = require("ytdl-core");
```
taahamahdi, line 1: This file can be removed

taahamahdi (changes requested):

taahamahdi (approved):

PR #1615, opened 2023-03-17
Strict mode: strictPropertyInitialization

Brainicism:


src/database_context.ts
```diff
@@ -37,7 +37,6 @@ export class DatabaseContext {
     public agnostic: Knex;
 
     constructor() {
-        if ([EnvType.CI].includes(process.env.NODE_ENV as EnvType)) return;
```
taahamahdi, line 40: Do we no longer need this early return for CI?

Brainicism: This was introduced before we seeded the kpop_videos database during the test setup, so trying to create a connection to that db would fail.

taahamahdi (approved):

PR #1639, opened 2023-04-15
Zero-downtime Docker container upgrade

Brainicism:


src/kmq.ts
```diff
@@ -167,24 +170,40 @@ function registerProcessEvents(fleet: Fleet): void {
 
         fleet.on("ready", async () => {
             logger.info("All shards have connected.");
+
+            // check if current instance is a standby currently being spun up, assign ready state
+            // as all shards are now ready
+            if (await pathExists(STANDBY_COOKIE)) {
+                await fs.promises.writeFile(STANDBY_COOKIE, "ready");
+
+                // wait for upgrade workflow to promote to primary by deleting standby cookiew
```
taahamahdi, line 179:
```suggestion
                // wait for upgrade workflow to promote to primary by deleting standby cookie
```

package.json
```diff
@@ -22,7 +22,7 @@
         "docker-compose-dev": "cd docker/ && NODE_ENV=development docker-compose up  --exit-code-from kmq --build kmq",
         "docker-compose-dryrun": "cd docker/ && NODE_ENV=dry-run docker-compose up --exit-code-from kmq --build kmq",
         "docker-build": "cd docker/ && DOCKER_BUILDKIT=1 docker build -f kmq/Dockerfile -t ghcr.io/brainicism/kmq_discord:latest ../",
-        "docker-run": ". ./.env && docker run --restart unless-stopped -d -v ${SONG_DOWNLOAD_DIR}:${SONG_DOWNLOAD_DIR} -v ${PWD}/data:/app/data -v ${PWD}/.env:/app/.env -v ${PWD}/logs:/app/logs -v ${PWD}/sql_dumps/kmq_backup:/app/sql_dumps/kmq_backup --env NODE_ENV=production --network=host --name ${APP_NAME} ${IMAGE_NAME:=ghcr.io/brainicism/kmq_discord:latest}",
```
taahamahdi, line 25: Should NODE_ENV be production here

src/scripts/announce-restart.ts
```diff
@@ -36,6 +36,10 @@ async function abortRestart(): Promise<void> {
     );
 }
 
+const delay = (time: number): Promise<void> =>
```
taahamahdi, line 39: We already have a delay function, do we need this here as well

taahamahdi (changes requested):

src/scripts/announce-restart.ts
Brainicism: Importing anything from `utils` has a dependency on our internal logger. Our internal logger has issues when writing to logs both from Docker and non-Docker scenarios, because Docker writes to the file as root. Would rather just avoid that for now and fix it later

taahamahdi (approved):

PR #1686, opened 2023-06-28
Fix bug where guess time would get overwritten

taahamahdi:
Keep track of all guesses. For hidden games, use the most recent guess. For non-hidden games, use the oldest correct guess.

### DIFF
--- a/src/structures/game_round.ts
+++ b/src/structures/game_round.ts
@@ -467,6 +467,7 @@ export default class GameRound extends Round {
                         const mostRecentGuessTime = Number(
                             Object.keys(x[1]).pop()
                         );
+
                         const mostRecentGuess = Object.values(x[1]).pop()!;
                         return [
                             playerID,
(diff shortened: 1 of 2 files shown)


src/structures/game_round.ts
```diff
@@ -505,6 +529,22 @@ export default class GameRound extends Round {
...
+            const playerIDToEarliestCorrectTimestamp: {
+                [playerID: string]: number;
+            } = {};
+
+            for (const [playerID, guesses] of Object.entries(this.guesses)) {
+                const earliestGuess = guesses
+                    .sort((a, b) => a.createdAt - b.createdAt)
+                    .find((x) => x.correct);
+
+                if (earliestGuess) {
```
Brainicism, line 541: When would this be null?

Brainicism (changes requested):

Brainicism (approved):

PR #1690, opened 2023-07-01
Prevent stale /spotify metadata

taahamahdi:
* Store only playlistID in game options and game option presets (script needs to be run)
* Cache queried songs and metadata (flushed daily, invalidated on snapshotID change)
* Show spotify thumbnail in /options
* Move MatchedPlaylist, PlaylistMetadata into interface files

### DIFF
--- a/src/helpers/spotify_manager.ts
+++ b/src/helpers/spotify_manager.ts
@@ -17,26 +17,13 @@ import dbContext from "../database_context";
 import fs from "fs";
 import path from "path";
 import type { AxiosResponse } from "axios";
+import type { MatchedPlaylist } from "../interfaces/matched_playlist";
 import type QueriedSong from "../interfaces/queried_song";
 import type SpotifyTrack from "../interfaces/spotify_track";
 
 const logger = new IPCLogger("spotify_manager");
 
 const BASE_URL = "https://api.spotify.com/v1";
-
-export interface PlaylistMetadata {
-    playlistID: string;
-    playlistName: string;
-    playlistLength: number;
-    matchedSongsLength: number;
-    thumbnailUrl?: string;
-}
-
-export interface MatchedPlaylist {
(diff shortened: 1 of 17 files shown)


src/scripts/spotify-playlist-metadata-to-id.ts
```diff
@@ -0,0 +1,92 @@
...
+                .values(playlistIDs)
+                .execute();
+
+            await dbContext.kmq
+                .deleteFrom("game_options")
+                .where("option_name", "=", "spotifyPlaylistMetadata")
+                .execute();
+        }
+
+        {
```
Brainicism, line 48: Is this codeblock duplicated?

src/helpers/spotify_manager.ts
```diff
@@ -144,104 +137,111 @@ export default class SpotifyManager {
...
         matchedSongs = _.uniqBy(matchedSongs, "youtubeLink");
+        const metadata = {
+            playlistID,
+            playlistLength: spotifyMetadata.songCount,
+            playlistName: spotifyMetadata.playlistName,
+            matchedSongsLength: matchedSongs.length,
+            thumbnailUrl: spotifyMetadata.thumbnailUrl as string,
+        };
+
+        State.cachedSpotifyPlaylists[playlistID] = {
```
Brainicism, line 235: nit: we're updating this cache even if the entry was originally retrieved from cache

Brainicism (changes requested):

src/scripts/spotify-playlist-metadata-to-id.ts
taahamahdi: One is for options, the other is for presets. Putting in a for loop with those 2 doesn't work well since there are fields that exist in one but not the other

Brainicism (dismissed):

Brainicism (approved):

PR #1692, opened 2023-07-03
Chance to show last news entry on /play

taahamahdi:
Screenshot 2023-07-03 at 2 27 05 AM


src/commands/game_commands/play.ts
```diff
@@ -167,12 +174,32 @@ export async function sendBeginGameSessionMessage(
         );
     }
 
+    const additionalPayloads = [optionsEmbedPayload];
+
+    let newsData = "";
+    try {
+        newsData = (await fs.promises.readFile(DataFiles.NEWS)).toString();
+        newsData = newsData.split("\n\n")[0];
```
Brainicism, line 182: We should only be showing recent news entries, not just the latest

Brainicism (approved):

Brainicism (approved):

PR #1716, opened 2023-07-11
Send lightning emoji on quick guess

taahamahdi:
Screenshot 2023-07-11 at 4 02 22 PM

### DIFF
--- a/src/commands/game_commands/exp.ts
+++ b/src/commands/game_commands/exp.ts
@@ -4,6 +4,8 @@ import {
     GUESS_STREAK_THRESHOLD,
     KmqImages,
     PARTICIPANT_MODIFIER_MAX_PARTICIPANTS,
+    QUICK_GUESS_EMOJI,
+    QUICK_GUESS_MS,
 } from "../../constants";
 import { IPCLogger } from "../../logger";
 import {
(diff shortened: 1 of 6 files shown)


src/commands/game_commands/exp.ts
```diff
@@ -418,13 +420,13 @@ export default class ExpCommand implements BaseCommand {
                 "command.exp.explanation.streak"
             )}:\` ${ExpBonusModifierValues[
                 ExpBonusModifier.GUESS_STREAK
-            ].toFixed(2)}x 📈`,
+            ].toFixed(2)}x 🔥`,
```
Brainicism, line 423: These changes are inconsistent with the other modifiers

src/structures/game_round.ts
```diff
@@ -530,14 +532,15 @@ export default class GameRound extends Round {
                     ? ` (+${friendlyFormattedNumber(playerResult.expGain)} EXP)`
                     : "";
 
+                const timeToGuess = durationSeconds(this.startedAt, createdAt);
```
Brainicism, line 535: Don't we already calculate guess times somewhere else? We should re-use so that there's only one source of truth

Brainicism (changes requested):

src/structures/game_round.ts
taahamahdi: `createdAt` is the most recent guess we have, from above. A similar variable `earliestGuess` exists for the non-hidden code path, and that is used below

Brainicism (approved):

PR #1717, opened 2023-07-11
Add progress bar for /spotify and /profile

taahamahdi:
* Also increases playlist parse time from 15s to 30s

### DIFF
--- a/src/helpers/spotify_manager.ts
+++ b/src/helpers/spotify_manager.ts
@@ -220,48 +220,54 @@ export default class SpotifyManager {
                 });
             }, 2000);
 
-            const songMatchStartTime = Date.now();
-            for await (const queryOutput of asyncPool(
-                4,
-                spotifySongs,
-                (x: SpotifyTrack) =>
-                    this.generateSongMatchingPromise(x, isPremium)
-            )) {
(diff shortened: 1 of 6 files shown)


src/helpers/spotify_manager.ts
```diff
@@ -210,6 +261,19 @@ export default class SpotifyManager {
                 }ms.`
             );
 
+            clearInterval(updateParsing);
```
Brainicism, line 264: What happens if an exception occurs before we reach here?

src/helpers/utils.ts
```diff
@@ -607,3 +607,24 @@ export function durationSeconds(startTime: number, endTime: number): number {
...
+ */
+export function visualProgressBar(
+    current: number,
+    total: number,
+    barLength = 10
+): string {
+    // Ensure the ratio is between 0 and 1
+    const ratio = Math.max(0, Math.min(1, current / total));
+
+    const completedBlocks = Math.round(ratio * barLength);
```
Brainicism, line 626: probably floor instead of round here

Brainicism (changes requested):

Brainicism (approved):

PR #1729, opened 2023-07-16
Show all news updates from past 30 days

Brainicism:


src/commands/game_commands/play.ts
```diff
@@ -176,36 +176,67 @@ export async function sendBeginGameSessionMessage(
...
-        const daysSinceUpdate = durationDays(
-            mostRecentUpdate.getTime(),
-            Date.now()
-        );
+    if (newsFileContent) {
+        const staleUpdateThreshold = 30;
+        const newsData: Array<{ updateTime: Date; entry: string }> =
+            newsFileContent
+                .split("\n\n")
+                .map((x) => x)
```
taahamahdi, line 193: This map seems redundant

taahamahdi (approved):

src/commands/game_commands/play.ts
Brainicism: It removes empty strings

Brainicism: woops it should have been a filter

taahamahdi (approved):

PR #1740, opened 2023-08-29
Fix bug where long embed titles weren't truncated

taahamahdi:


src/structures/session.ts
```diff
@@ -913,7 +913,7 @@ export default abstract class Session {
         const songAndArtist = `"${getLocalizedSongName(
             round.song,
             locale
-        )}" - ${getLocalizedArtistName(round.song, locale)}`;
+        )}" - ${getLocalizedArtistName(round.song, locale)}`.substring(0, 256);
```
Brainicism, line 916: Do we really want to show 256 characters?

Brainicism (changes requested):

src/helpers/utils.ts
```diff
@@ -628,3 +628,16 @@ export function visualProgressBar(
...
+ * @param input - the input string
+ * @param maxLength - the maximum length of the string
+ * @returns the input string, truncated to the given length
+ */
+export function truncatedString(input: string, maxLength: number): string {
+    if (input.length <= maxLength) {
+        return input;
+    }
+
+    return `${input.substring(0, maxLength - 3)}...`;
```
Brainicism, line 642: This `-3` padding should not exist in a generic function like this

Brainicism (approved):

PR #1756, opened 2023-10-23
Add 10-min job to cleanup inactive listening sessions

taahamahdi:


src/helpers/game_utils.ts
```diff
@@ -149,6 +149,28 @@ export async function cleanupInactiveGameSessions(): Promise<void> {
...
+            const listeningSession = listeningSessions[guildID];
+            if (listeningSession.getVoiceMembers().length === 0) {
+                await listeningSession.endSession("Empty listening session");
+                inactiveSessions++;
+            }
+        })
+    );
+
+    if (inactiveSessions > 0) {
+        logger.info(
```
Brainicism, line 168: Lingering listening sessions shouldn't be happening. Should emit this as a warning/error so we can figure out how its happening

Brainicism (approved):

PR #1762, opened 2023-10-27
Support Spotify shorthand playlist URL

### LINKED ISSUE #1761: Handle Spotify shortlinks
Either support them, or include message on the proper format to use

taahamahdi:
Closes #1761 

There isn't an API endpoint to convert this to the full URL, so making a web request and parsing out the URL is the only way to get the playlist ID


Brainicism (dismissed):

src/commands/game_options/spotify.ts
```diff
@@ -225,139 +242,174 @@ export default class SpotifyCommand implements BaseCommand {
...
-            await guildPreference.setSpotifyPlaylistID(playlistID);
-
-            await LimitCommand.updateOption(
-                messageContext,
-                0,
-                matchedPlaylist.metadata.matchedSongsLength,
-                undefined,
-                false,
-            );
+        if (playlistID.includes("?si=")) {
```
Brainicism, line 310: Can we not use a regex to extract the playlist id instead?

taahamahdi: This check will only apply to the normal Spotify url. I'll move it into there

Brainicism (approved):

PR #1794, opened 2023-12-27
Add AI-based news command

taahamahdi:
Screenshot 2024-01-04 at 2 38 41 AM


src/helpers/management_utils.ts
```diff
@@ -442,11 +451,80 @@ async function reloadBanData(): Promise<void> {
     State.bannedPlayers = new Set(bannedPlayers);
 }
 
+async function registerNewsSubscriptions(): Promise<void> {
```
Brainicism, line 454: nit: `register` sounds like we're persisting a new subscription?

src/helpers/reddit_client.ts
```diff
@@ -11,6 +11,25 @@ export interface KpopNewsRedditPost {
     flair: string;
 }
 
+// TODO filter out [Song Cover]", "[Live]", "[Variety]", "[Behind-The-Scenes]", "[CF]", "[Audio]", "[Interview]", "[Dance Challenge]", "[Meta]"
```
Brainicism, line 14: remove TODO

src/helpers/reddit_client.ts
```diff
@@ -56,4 +75,56 @@ export class RedditClient {
...
+    async getTopDayPosts(): Promise<Array<KpopNewsRedditPost>> {
+        try {
+            const matchingPosts = await this.client.search({
+                subreddit: "kpop",
+                query: generateFilteredQuery(),
+                sort: "top",
+                time: "day",
+            });
+
+            const popularPosts = matchingPosts.filter((x) => x.score > 100);
```
Brainicism, line 88: Is karma > 100 too high for daily posts?

src/helpers/reddit_client.ts
```diff
@@ -56,4 +75,56 @@ export class RedditClient {
...
+            }));
+        } catch (e) {
+            logger.error(
+                `Failed to fetch getTopDayPosts(). e = ${JSON.stringify(e)}`,
+            );
+            return [];
+        }
+    }
+
+    async getTopWeekPosts(): Promise<Array<KpopNewsRedditPost>> {
```
Brainicism, line 105: could we share logic between daily/weekly/monthly somehow? most of it seems teh same

src/fact_generator.ts
```diff
@@ -100,9 +104,9 @@ interface GaonWeeklyEntry {
 /**
  * Reloads the fact cache
  */
-export async function reloadFactCache(): Promise<void> {
+export function reloadFactCache(): void {
     logger.info("Regenerating fact cache...");
-    await generateFacts();
+    generateFacts();
```
Brainicism, line 109: Why rid of await

src/fact_generator.ts
```diff
@@ -125,17 +129,19 @@ async function resolveFactPromises(
     return resolvedPromises.map((x) => x["value"]);
 }
 
-async function generateFacts(): Promise<void> {
+function generateFacts(): void {
     logger.info("Generating Reddit news facts...");
-    const newsFacts = await resolveFactPromises([redditKpopNews()]);
```
Brainicism, line 130: This is agnostic of locale type, why we moving it into the loop

src/commands/misc_commands/news.ts
```diff
@@ -0,0 +1,443 @@
...
+                messageContext.guildID,
+                "command.news.subscribe.title",
+            ),
+            description: i18n.translate(
+                messageContext.guildID,
+                "command.news.subscribe.description",
+            ),
+        });
+
+        await this.sendNews(messageContext, range, interaction);
```
Brainicism, line 378: Why do we send news on subscription?

src/commands/misc_commands/news.ts
```diff
@@ -0,0 +1,443 @@
...
+                    "command.news.unsubscribe.description",
+                ),
+            },
+            false,
+            undefined,
+            [],
+            interaction,
+        );
+
+        schedule.cancelJob(
```
Brainicism, line 404: We should check if the guild has subscribed to begin with, otherwise they might think this commmand is doing nothing

Brainicism: same with in `subscribeNews`

src/commands/misc_commands/news.ts
```diff
@@ -0,0 +1,443 @@
...
+    };
+
+    slashCommands = (): Array<
+        DefaultSlashCommand | Eris.ChatInputApplicationCommandStructure
+    > => [
+        {
+            type: Eris.Constants.ApplicationCommandTypes.CHAT_INPUT,
+            options: [
+                {
+                    name: Action.GET,
```
Brainicism, line 92: `/news get` isnt an intuitive command for the user, especially for a command that might be used frequently. we can just do `/news [daily/weekly/month]`

src/commands/misc_commands/news.ts
```diff
@@ -0,0 +1,443 @@
...
+        const locale = State.getGuildLocale(messageContext.guildID);
+
+        let summary: string;
+        try {
+            summary = State.news[range][locale];
+        } catch (err) {
+            // Failed to generate news since startup
+            logger.error(
+                `${getDebugLogHeader(messageContext)} | Error sending news: ${
+                    err.message
```
Brainicism, line 263: `err` wouldnt contain any useful information would it? it would just be a null pointer reference

src/helpers/gemini_client.ts
```diff
@@ -0,0 +1,102 @@
...
+    posts: Array<Object>,
+    interval: Interval,
+    locale: LocaleType,
+): string => {
+    let prompt = `You are Kimiqo, a friendly 23 year old K-pop enthusiast who follows the latest updates in K-pop. You are giving an update on the latest happenings in K-pop for a game called KMQ (K-pop Music Quiz). You will be given a string delimited by |, where each column is: the date, the type of post, and the title of the post. The posts are sorted by priority/significance.  Make sure to address the message to KMQ fans/players, and mention who you are. Limit your response to 250 words, even if it means you have to ignore some of the data.
+
+    Summarize ${interval} in K-pop in paragraph form (multiple paragraphs if needed), from the POV of an excited and preppy K-pop fan. Add a lot of personality to the summary. Use emojis where appropriate.`;
+
+    if (locale !== LocaleType.EN) {
+        prompt += ` Respond in the locale ${locale}.`;
```
Brainicism, line 25:
```suggestion
        prompt += ` Respond in the locale: ${locale}.`;
```

src/helpers/gemini_client.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    return `${prompt} Data is below.\n:\n ${posts.join("\n")}`;
+};
+
+export default class GeminiClient {
+    private client: GenerativeModel;
+
+    constructor() {
+        const genAI = new GoogleGenerativeAI(process.env.GEMINI_API_KEY!);
```
Brainicism, line 35: how does this class/feature behaved if we forget to give it an API key?

src/helpers/gemini_client.ts
```diff
@@ -0,0 +1,102 @@
...
+            );
+
+            const generatedContent = await this.client.generateContent(
+                getNewsPrompt(topDayPosts, Interval.DAY, locale),
+            );
+
+            const text = generatedContent.response.text();
+            if (text === "") {
+                logger.warn(
+                    `Failed to generate text for getDailyPostSummary(). generatedContent = ${JSON.stringify(
```
Brainicism, line 53: We should log (or persist into `data/failed_prompts`) the failing prompts

src/helpers/gemini_client.ts
```diff
@@ -0,0 +1,102 @@
...
+            logger.error(
+                `Failed to fetch getDailyPostSummary(). e = ${JSON.stringify(
+                    e,
+                )}`,
+            );
+            return "";
+        }
+    }
+
+    async getWeeklyPostSummary(locale: LocaleType): Promise<string> {
```
Brainicism, line 70: Can we consolidate these two functions somehow

src/helpers/gemini_client.ts
```diff
@@ -0,0 +1,102 @@
...
+            const text = generatedContent.response.text();
+            if (text === "") {
+                logger.warn(
+                    `Failed to generate text for getDailyPostSummary(). generatedContent = ${JSON.stringify(
+                        generatedContent,
+                    )}. topDayPosts = ${topDayPosts}`,
+                );
+            }
+
+            return text;
```
Brainicism, line 59: We should add checks (+potential retry) if the output isn't the desired size.

Brainicism (changes requested):

src/helpers/reddit_client.ts
Brainicism: Also, do we want to limit the number of posts?

taahamahdi: Currently, there are 20 posts above 100 in the past 24 hours. I can set a limit to 25, which is the first page of r/kpop on old reddit

src/commands/misc_commands/news.ts
taahamahdi: I don't think it's possible to mix a regular command with subcommands, is there another word to replace `get` with?

src/commands/misc_commands/news.ts
taahamahdi: That's right, I can take this log out and just print an error message

src/helpers/gemini_client.ts
taahamahdi: I've updated the behavior so it lets the user know to try again later

Brainicism (dismissed):

Brainicism (approved):

Brainicism (approved):

Brainicism (approved):

PR #1805, opened 2024-01-09
Synchronously generate news

taahamahdi:
Attempt to avoid the gemini rate limit


kambojk (approved):
i dont know whats going on but thats cool i guess

PR #1833, opened 2024-01-26
Fix bug where pre-round guesses were accepted

taahamahdi:


src/structures/round.ts
```diff
@@ -20,7 +20,7 @@ export default abstract class Round {
     public readonly artistAliases: string[];
 
     /** Timestamp of the creation of the Round in epoch milliseconds */
-    public readonly startedAt: number;
+    public startedAt: number;
```
Brainicism, line 23: If we need seperate variable to track when audio starts playing, should create it.

src/structures/round.ts
```diff
@@ -57,6 +60,7 @@ export default abstract class Round {
             (x) => State.aliases.artist[x] || [],
         );
         this.startedAt = Date.now();
+        this.songStartedAt = Date.now();
```
Brainicism, line 63: Should be null if song did not start yet

Brainicism (changes requested):

src/structures/game_round.ts
```diff
@@ -346,9 +349,16 @@ export default class GameRound extends Round {
             typosAllowed,
         );
 
+        if (this.songStartedAt === null) {
+            logger.error(
+                `uid: ${playerID}, guess: ${guess}, createdAt: ${createdAt} | songStartedAt is unexpectedly null`,
+            );
+            this.songStartedAt = this.startedAt;
```
Brainicism, line 356: Why is this a valid guess if the song hasn't started?

src/structures/game_session.ts
```diff
@@ -352,7 +352,16 @@ export default class GameSession extends Session {
...
         round.interactionMarkAnswers(guessResult.correctGuessers?.length ?? 0);
 
-        const timePlayed = Date.now() - round.startedAt;
+        if (round.songStartedAt === null) {
+            logger.error(
+                `${getDebugLogHeader(
+                    messageContext,
+                )} | songStartedAt is unexpectedly null`,
+            );
+            round.songStartedAt = round.startedAt;
```
Brainicism, line 361: Same here

src/structures/game_round.ts
taahamahdi: I needed a default value to put into these variables since it can be null. With the current logic, we shouldn't be able to enter this branch since when `songStartedAt` is null, `storeGuess` never gets called. Should I replace this (and the other commented section) as `songStartedAt!` on the variable instead?

Brainicism: You don't need a default value, they are separate variables for a reason. Just need to handle the null properly, in this case, if it's null then its clearly not a valid guess that we should store

Brainicism (approved):

PR #1871, opened 2024-02-10
Add i18n-ai-translate GH action

taahamahdi:


.github/workflows/lint-translations.yml
```diff
@@ -0,0 +1,58 @@
...
+        ref: ${{ github.head_ref }}
+        fetch-depth: 0
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '18'
+
+    - name: Install Dependencies
+      run: yarn add i18n-ai-translate
```
Brainicism, line 26: nit: If we install globally instead, it won't try to check for intercompatibility with our existing modules, which should save a lot of time

package.json
```diff
@@ -67,6 +67,7 @@
         "fastify": "^4.26.0",
         "fluent-ffmpeg": "^2.1.2",
         "googleapis": "^131.0.0",
+        "i18n-ai-translate": "^1.2.0",
```
Brainicism, line 70: Do we need this as part of our dependencies if we're only using/installing it during this GH action?

Brainicism: If yes, it should be a dev dependency

Brainicism (approved):

Brainicism (dismissed):

Brainicism (approved):

PR #1894, opened 2024-02-21
Change hidden to answer option

taahamahdi:


src/test/ci/command_prechecks.test.ts
```diff
@@ -494,187 +495,19 @@ describe("command prechecks", () => {
...
-                });
-            });
-        });
-    });
-
     describe("timerHiddenPrecheck", () => {
         const GUILD_ID = "123";
 
-        describe("hidden game session", () => {
+        describe("hidden answer type", () => {
```
Brainicism, line 501: This test is no longer relevant

Brainicism (changes requested):

src/test/ci/command_prechecks.test.ts
taahamahdi: I think it still is, we prevent disabling the timer when hidden is enabled since it requires it

Brainicism (approved):

PR #1902, opened 2024-02-24
Added emojis for language and remix tags to lookup.

remi1111:


src/helpers/game_utils.ts
```diff
@@ -563,6 +563,34 @@ export function getLocalizedArtistName(
     return song.hangulArtistName || song.artistName;
 }
 
+export function getTagsFromSong(daisukiEntry: { tags: string | null }): string {
+    let tag_text: string = "";
```
Brainicism, line 567: nit: camelcase for local variables

src/helpers/game_utils.ts
```diff
@@ -563,6 +563,34 @@ export function getLocalizedArtistName(
...
+    if (daisukiEntry.tags === null) {
+        return tag_text;
+    } else {
+        if (daisukiEntry.tags.indexOf("e") != -1) {
```
Brainicism, line 571: You can use `.includes` to check for existence of char in string

src/helpers/game_utils.ts
```diff
@@ -563,6 +563,34 @@ export function getLocalizedArtistName(
...
+        if (daisukiEntry.tags.indexOf("s") != -1) {
+            tag_text += " :flag_es:"; // Spanish
+        }
+        if (daisukiEntry.tags.indexOf("l") != -1) {
+            tag_text += " :globe_with_meridians:"; // Other Language
+        }
+        if (daisukiEntry.tags.indexOf("x") != -1) {
+            tag_text += " :arrows_counterclockwise:"; // Remix
+        }
+    }
```
Brainicism, line 589: If you wanted to be fancier, you can create a dict of tag to emoji, and iterate over that instead of creating individual if statements here.

src/helpers/game_utils.ts
```diff
@@ -563,6 +563,34 @@ export function getLocalizedArtistName(
     return song.hangulArtistName || song.artistName;
 }
 
+export function getTagsFromSong(daisukiEntry: { tags: string | null }): string {
```
Brainicism, line 566: Name is slightly misleading here? It takes in tags as param and outputs emojis

Brainicism (changes requested):

src/helpers/game_utils.ts
```diff
@@ -563,6 +563,34 @@ export function getLocalizedArtistName(
...
+    let tag_text: string = "";
+    if (daisukiEntry.tags === null) {
+        return tag_text;
+    } else {
```
Brainicism, line 570: Would prefer early return here to be consistent with codebase

src/commands/game_commands/lookup.ts
```diff
@@ -290,7 +294,7 @@ async function lookupByYoutubeID(
     sendInfoMessage(
         messageContext,
         {
-            title: `"${songName}" - ${artistName}`,
+            title: `"${songName}" - ${artistName}${tags}`,
```
taahamahdi, line 297: I'd recommend a space in between the artist name and the first emoji

src/commands/game_commands/lookup.ts
```diff
@@ -114,6 +116,7 @@ async function lookupByYoutubeID(
     let artistName: string;
     const songAliases: string[] = [];
     const artistAliases: string[] = [];
+    const tags: string = getEmojisFromSongTags(daisukiEntry);
```
Brainicism, line 119: nit: variable name needed to be updated
nit: this variable is only used in one place, can we just call `getEmojisFromSongTags()` directly where it's used?

Brainicism (approved):

PR #1920, opened 2024-03-02
Cache app command IDs

Brainicism:
https://discord.com/developers/docs/interactions/application-commands#updating-and-deleting-a-command
Application commands do not change their IDs, and command updates to the same type/scope are treated as upserts. 

Cache and load from cache upon startup incase of command update timing out.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -240,6 +240,7 @@ export const DataFiles = {
         __dirname,
         "../data/feature_switch_config.json",
     ),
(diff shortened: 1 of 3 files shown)


taahamahdi (approved):

taahamahdi:
Alternatively since we know that the get slash commands endpoint hasn't been having issues, we can fetch IDs by default and only reload on demand through ,appcommands

Brainicism:
> Alternatively since we know that the get slash commands endpoint hasn't been having issues, we can fetch IDs by default and only reload on demand through ,appcommands

It is better to rely on automation than to remember to do things because human memory is bad

PR #1967, opened 2024-03-17
Update i18n-ai-translate, use ChatGPT

taahamahdi:


src/ci_checks/i18n-ai-translate.sh
```diff
@@ -7,7 +7,7 @@ cp i18n/en.json i18n/en-latest.json
 git checkout $merge_base -- i18n/en.json
 if ! diff i18n/en-latest.json i18n/en.json; then
     echo "Updating translations..."
-    npx i18n-ai-translate diff -b i18n/en.json -a i18n/en-latest.json -l "English" --verbose
+    npx i18n-ai-translate diff -b i18n/en.json -a i18n/en-latest.json -l "English" --verbose --engine chatgpt
```
Brainicism, line 10: does this default to gpt3.5 or 4?

Brainicism: Don’t think this is something that should have a default, especially if both are available to users and have different costs

taahamahdi: Is it something that should be fixed on the library side or something here

Brainicism: library, but a result, here as well

Brainicism (approved):

PR #1980, opened 2024-03-20
Remove state dependency from game utils

taahamahdi:


src/test/ci/game_utils.test.ts
```diff
@@ -41,9 +41,10 @@ describe("game utils", () => {
             describe("collabs", () => {
                 it("should return the group and any collabs they are a part of in matchedGroups", async () => {
                     const artistName = "IU";
-                    const matchResults = await getMatchingGroupNames([
-                        artistName,
-                    ]);
+                    const matchResults = await getMatchingGroupNames(
+                        State.aliases.artist,
```
Brainicism, line 45: We don't need any dependency on `State` in these tests anymore, we can just use local variable

Brainicism (changes requested):

src/test/ci/game_utils.test.ts
taahamahdi: This may be done easier when I remove state from management_utils, is it okay. to leave it in for now

Brainicism (approved):

PR #1984, opened 2024-03-21
Add test runner and test suite for emulating user commands

Brainicism:


src/test/end-to-test/test-runner-bot.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    console.log(
+        `=====================STAGE ${CURRENT_STAGE}===================`,
+    );
+
+    const stageData = BASIC_OPTIONS_TEST_SUITE.tests[CURRENT_STAGE]!;
+    const command = stageData.command;
+    if (TEST_SUITE.resetEachStage) {
+        await sendCommand(",reset");
```
taahamahdi, line 86: Should this use the prefix

src/test/end-to-test/test_suites/basic_options_test.ts
```diff
@@ -0,0 +1,237 @@
+import type ParsedGameOptionValues from "../parsed_game_options_value";
+import type TestSuite from "./test_suite";
+
+const BASIC_OPTIONS_TEST_SUITE: TestSuite = {
+    tests: [
+        {
+            command: ",limit 2",
```
taahamahdi, line 7: These should use bot prefix

taahamahdi (approved):
Some suggestions but looks good

src/helpers/discord_utils.ts
```diff
@@ -439,6 +439,21 @@ export async function sendMessage(
     authorID?: string,
     interaction?: Eris.ComponentInteraction | Eris.CommandInteraction,
 ): Promise<Eris.Message | null> {
+    if (messageContent.messageReference) {
+        const message: Message<TextChannel> | undefined = await (
+            State.client.getChannel(textChannelID!) as
+                | Eris.TextChannel
+                | undefined
+        )?.getMessage(messageContent.messageReference.messageID);
```
taahamahdi, line 447: Will this lookup slow down non-test send message calls

Brainicism: These are cached

src/test/end-to-test/test_suites/basic_options_test.ts
Brainicism: They are replaced with bot prefix in the runner code

src/test/end-to-test/test-runner-bot.ts
Brainicism: It's replaced in sendCommand

PR #1987, opened 2024-03-22
Add end-to-end gameplay test

Brainicism:


src/test/end-to-end-tests/test_suites/play_test.ts
```diff
@@ -0,0 +1,228 @@
...
+                parsedGameOptions?: ParsedGameOptionValues,
+                client?: Eris.Client,
+            ) =>
+                parsedGameOptions!["limit set top"]!.value.startsWith("6 /") &&
+                parsedGameOptions!["limit set top"]!.updated,
+            expectedResponseType: KmqResponseType.GAME_OPTIONS_RESPONSE,
+            requiresVoiceConnection: true,
+        },
+        {
+            command: ",shuffle popularity", // deterministic ordering so we can figure out how to guess
```
taahamahdi, line 37: Will this ordering break once one of these six songs changes rank? Should we query the DB to get the song/artist names instead

src/test/end-to-end-tests/test_suites/play_test.ts
```diff
@@ -0,0 +1,228 @@
...
+                title: string,
+                description: string,
+                parsedGameOptions?: ParsedGameOptionValues,
+                client?: Eris.Client,
+            ) => true,
+            expectedResponseType: KmqResponseType.RAW,
+            requiresVoiceConnection: true,
+        },
+        {
+            command: "gangnma style", // gangnam style - psy, lost guess streak, unique songs reset, allow typo
```
taahamahdi, line 175: Do we already have a command that sets it to typingtypos elsewhere? Is it a part of the other test file?

Brainicism: It's the default

src/test/end-to-end-tests/test_suites/play_test.ts
Brainicism: Yes but would rather keep the test runner agnostic of anything internal, it should be as close to a regular player as possible. 

Many of our tests break when the database changes in certain ways. We use a cached version of the database so it doesn't break unless we update it, and fixing things doesn't take much effort

taahamahdi: Good point, let's leave it as is then

taahamahdi (approved):

taahamahdi (approved):

PR #2012, opened 2024-03-27
Collabs can be added to group filters.

remi1111:
- Collabs can now be added to group, include and exclude

- Add, Remove and Groups now handle collabs spawned from artists correctly.

- All collabs no longer show on group, include or exclude unless directly added by the player.

- Duplicate entries no longer show up.

### DIFF
--- a/src/commands/game_options/remove.ts
+++ b/src/commands/game_options/remove.ts
@@ -209,6 +209,8 @@ export default class RemoveCommand implements BaseCommand {
         const remainingGroups = currentMatchedArtists.filter(
             (group) => !matchedGroups.some((x) => x.id === group.id),
         );
+
+        // Rematch the remaining groups to make sure collabs are not removed by accident.
         const groups = await getMatchingGroupNames(
             State.aliases.artist,
             remainingGroups.map((x) => x.name),
--- a/src/commands/game_options/remove.ts
+++ b/src/commands/game_options/remove.ts
@@ -124,13 +124,13 @@ export default class RemoveCommand implements BaseCommand {
             return;
         }
 
+        currentMatchedArtists = currentMatchedArtists.filter((groups) =>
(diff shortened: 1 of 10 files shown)


src/test/unit_tests/ci/game_utils.test.ts
```diff
@@ -327,23 +327,23 @@ describe("game utils", () => {
...
+                        it("should match group", async () => {
+                            const artistName = "PSY + Hyuna";
+
                             const matchResults = await getMatchingGroupNames(
                                 State.aliases.artist,
-                                [nonMatchArtist],
+                                [artistName],
                             );
 
                             assert.deepStrictEqual(
```
Brainicism, line 339: Can you add an assert that checks that `addedByUser` is true? And another test that checks that its false if it was added through an artist instead?

src/interfaces/matched_artist.ts
```diff
@@ -1,5 +1,6 @@
 export default interface MatchedArtist {
     id: number;
     name: string;
+    added_by_user: string;
```
Brainicism, line 4: - Can this be in camelcase?
- Can this be boolean?

Brainicism (changes requested):

src/commands/game_options/remove.ts
```diff
@@ -206,24 +206,34 @@ export default class RemoveCommand implements BaseCommand {
             return;
         }
 
+        const remainingGroups = currentMatchedArtists.filter(
+            (group) => !matchedGroups.some((x) => x.id === group.id),
+        );
+
+        // Rematch the remaining groups to make sure collabs are not removed by accident.
```
Brainicism, line 213: Is this re-matching still needed if we know which artists were added explicitly by the user?

Brainicism (approved):

PR #2021, opened 2024-03-29
Add god mode for having every song available

Brainicism:
image


src/structures/session.ts
```diff
@@ -676,6 +680,24 @@ export default abstract class Session {
                 break;
         }
 
+        if (isGodMode) {
+            /*
+                오빤 강남스타일
```
taahamahdi, line 685: 🕺

taahamahdi (approved):

PR #2027, opened 2024-03-29
Clip game mode

taahamahdi:


src/commands/game_commands/play.ts
```diff
@@ -1153,6 +1181,28 @@ export default class PlayCommand implements BaseCommand {
             }
         }
 
+        if (
```
Brainicism, line 1184: Can you add explanation of whats happening here?

src/commands/game_commands/play.ts
```diff
@@ -1153,6 +1181,28 @@ export default class PlayCommand implements BaseCommand {
...
+            gameType === GameType.CLIP &&
+            (!guildPreference.gameOptions.guessTimeout ||
+                guildPreference.gameOptions.guessTimeout >
+                    GuessTimeoutCommand.TIMER_MAX_ACCEPTABLE_CLIP_BEFORE_RESET)
+        ) {
+            await guildPreference.setGuessTimeout(
+                GuessTimeoutCommand.TIMER_DEFAULT_VALUE,
```
Brainicism, line 1191: Why are we comparing with TIMER_MAX_ACCEPTABLE_CLIP_BEFORE_RESET but capping it to TIMER_DEFAULT_VALUE?

src/commands/game_commands/play.ts
```diff
@@ -1153,6 +1181,28 @@ export default class PlayCommand implements BaseCommand {
...
+            );
+        }
+
+        if (
+            gameType !== GameType.CLIP &&
+            guildPreference.isGuessTimeoutSet() &&
```
Brainicism, line 1197: Above you use `!guildPreference.gameOptions.guessTimeout`, here you use `guildPreference.isGuessTimeoutSet()`. Should be consistent here (probably leaning towards the second, since 0 is falsy)

Brainicism: Can we also split this conditional by gameType?
```ts
if (gameType === GameType.CLIP) {

    ...
} else{
    ...
}

Would make this a lot easier to read

File: src/commands/game_commands/play.ts [hunk 4]
```diff
@@ -1153,6 +1181,28 @@ export default class PlayCommand implements BaseCommand {
             }
         }
 
+        if (
+            gameType === GameType.CLIP &&
+            (!guildPreference.gameOptions.guessTimeout ||
+                guildPreference.gameOptions.guessTimeout >
+                    GuessTimeoutCommand.TIMER_MAX_ACCEPTABLE_CLIP_BEFORE_RESET)
+        ) {
+            await guildPreference.setGuessTimeout(
```
Brainicism, line 1190: What happens if they try to set the guess timeout after the game starts already?

src/commands/game_options/timer.ts
```diff
@@ -24,8 +26,11 @@ const logger = new IPCLogger("guessTimeout");
 
 // eslint-disable-next-line import/no-unused-modules
 export default class GuessTimeoutCommand implements BaseCommand {
-    static TIMER_MIN_VALUE = 2;
+    static TIMER_MIN_VALUE = 0.75;
```
Brainicism, line 29: So this is the min value for CLIP?
Maybe call it TIMER_MIN_VALUE_CLIP, and have another variable
TIMER_MIN_VALUE = min(TIMER_MIN_VALUE_CLIP, TIMER_MIN_VALUE_NON_CLIP) to be more explicit

src/commands/game_options/timer.ts
```diff
@@ -24,8 +26,11 @@ const logger = new IPCLogger("guessTimeout");
...
+    static TIMER_DEFAULT_VALUE = 2;
+    static TIMER_MIN_VALUE_NON_CLIP = 2;
     static TIMER_MAX_VALUE = 180;
+    static TIMER_MAX_ACCEPTABLE_CLIP_BEFORE_RESET = 5;
```
Brainicism, line 33: wat is this

src/commands/game_options/timer.ts
```diff
@@ -199,6 +204,39 @@ export default class GuessTimeoutCommand implements BaseCommand {
                 )} | Guess timeout disabled.`,
             );
         } else {
+            timer = Math.round(timer! * 100) / 100;
+            if (timer < GuessTimeoutCommand.TIMER_MIN_VALUE_NON_CLIP) {
+                if (session) {
+                    if (!session.isGameSession() || !session.isClipMode()) {
```
Brainicism, line 210: Doesn't `!session.isGameSession()` imply `!session.isClipMode()`?

src/commands/game_options/timer.ts
```diff
@@ -199,6 +204,39 @@ export default class GuessTimeoutCommand implements BaseCommand {
...
+                            {
+                                title: "command.timer.failure.tooLowTimer.title",
+                                description:
+                                    "command.timer.failure.tooLowTimer.description",
+                            },
+                            interaction,
+                        );
+                        return;
+                    }
+                } else {
```
Brainicism, line 222: It's hard to reason about what cases this 'else' is covering

src/enums/clip_action.ts
```diff
@@ -0,0 +1,6 @@
+enum ClipAction {
+    NEW_CLIP = "new_clip",
+    REPLAY = "replay_clip",
```
Brainicism, line 3: Do we still have a replay action?

src/helpers/validate.ts
```diff
@@ -101,8 +101,8 @@ export default async (
                     return false;
                 }
 
-                // parse as integer for now, might cause problems later?
-                const intArg = parseInt(arg, 10);
+                // parse as float for now, might cause problems later? :O
+                const intArg = parseFloat(arg);
```
Brainicism, line 105: Will this break anything? Also, rename the var

src/structures/clip_game_round.ts
```diff
@@ -0,0 +1,81 @@
+import { getMajorityCount } from "../helpers/discord_utils";
+import ClipAction from "../enums/clip_action";
+import GameRound from "./game_round";
+import type QueriedSong from "./queried_song";
+
+export default class ClipGameRound extends GameRound {
+    /** The Discord Guild ID */
+    public readonly guildID: string;
```
Brainicism, line 8: Should we just add this to `Round`?

src/structures/clip_game_round.ts
```diff
@@ -0,0 +1,81 @@
...
+     * Increments the replay count
+     */
+    incrementReplays(): void {
+        this.replays++;
+    }
+
+    /**
+     * @returns the number of replays
+     */
+    getReplays(): number {
```
Brainicism, line 70: getReplayCount to be more specific?

src/structures/game_round.ts
```diff
@@ -366,46 +367,60 @@ export default class GameRound extends Round {
...
-                            label += ` (${
-                                this.interactionIncorrectAnswerUUIDs[
-                                    z.custom_id
-                                ]
-                            })`;
-                            style = 4;
-                        }
-
+        const components: Eris.ActionRow[] = this.interactionComponents.map(
+            (x) => ({
```
Brainicism, line 371: can we rename these variables to be more specific (nested anonymous args is a bit confusing)
x => action row
y => component/button?

src/structures/game_round.ts
```diff
@@ -366,46 +367,60 @@ export default class GameRound extends Round {
...
+                type: 1,
+                components: x.components.map((y: Eris.InteractionButton) => {
+                    if (
+                        Object.values(ClipAction).includes(
+                            y.custom_id as ClipAction,
+                        )
+                    ) {
+                        // TODO heart here
                         return {
```
Brainicism, line 380: are we just trying to do a no-op here?

src/structures/game_session.ts
```diff
@@ -619,6 +612,62 @@ export default class GameSession extends Session {
         await super.updateOwner();
     }
 
+    async handleClipComponentInteraction(
+        round: GameRound,
+        interaction: Eris.ComponentInteraction<Eris.TextableChannel>,
+        messageContext: MessageContext,
+    ): Promise<boolean> {
+        if (
+            Date.now() - round.songStartedAt! <
```
Brainicism, line 621:
```
round.songStartedAt!
```
Is it possible for the clip button to be sent before the song starts and this is assigned a value? lets add a null check here just incase

Brainicism: Is this conditional just saying "cant request new clip" before the first replay?

src/typings/kmq_db.d.ts
```diff
@@ -219,13 +219,6 @@ export interface PlayerStats {
     songs_guessed: Generated<number>;
 }
 
-export interface PremiumUsers {
```
Brainicism, line 222: Not that it matters, but what happened here

src/constants.ts
```diff
@@ -62,6 +62,9 @@ export const DEFAULT_ADVANCED_SETTINGS = {
     [AdvancedCommandAction.SONG_START_DELAY]: 3,
 };
 
+export const CLIP_REPLAY_DELAY = 1000;
```
Brainicism, line 65: Are these used anywhere other than session?

src/structures/session.ts
```diff
@@ -664,17 +687,24 @@ export default abstract class Session {
...
-                seekLocation = 0;
-                break;
-            case SeekType.MIDDLE:
-                seekLocation = songDuration * (0.4 + 0.2 * Math.random());
-                break;
-            case SeekType.RANDOM:
-            default:
-                seekLocation = songDuration * (0.6 * Math.random());
-                break;
+        if (!round.songStartedAt) {
```
Brainicism, line 690: I am confused by this conditional. Are we assuming if its not CLIP mode, songStartedAt is non-null? If so, shouldnt we just check gameType != CLIP?

src/structures/session.ts
```diff
@@ -708,22 +738,29 @@ export default abstract class Session {
         this.connection.stopPlaying();
 
         try {
-            let inputArgs = ["-ss", seekLocation.toString()];
+            let inputArgs = ["-ss", seekLocation!.toString()];
```
Brainicism, line 741: Why is seekLocation nullable here?

src/structures/session.ts
```diff
@@ -708,22 +738,29 @@ export default abstract class Session {
...
                     songDuration,
                 );
 
                 inputArgs = ffmpegArgs.inputArgs;
                 encoderArgs = ffmpegArgs.encoderArgs;
             }
 
+            if (clipAction) {
+                const durationLimit =
+                    this.guildPreference.gameOptions.guessTimeout;
```
Brainicism, line 759: Guess timeout is being used as the amount of time the song plays? This doesn't seem right

src/structures/session.ts
```diff
@@ -737,6 +774,10 @@ export default abstract class Session {
             return false;
         }
 
+        if (this.isGameSession() && this.isClipMode()) {
```
Brainicism, line 777: Doesn't clip mode imply game session?

Brainicism: also cant we just set this on line 705

src/structures/session.ts
```diff
@@ -754,6 +795,19 @@ export default abstract class Session {
 
             this.stopGuessTimeout();
 
+            if (this.isGameSession() && this.isClipMode()) {
```
Brainicism, line 798: clipmode => game session?

Brainicism (changes requested):

src/structures/game_session.ts
```diff
@@ -1779,11 +1863,70 @@ export default class GameSession extends Session {
...
+                },
+            ],
+        });
+    }
+
+    private generateClipButton(): Eris.InteractionButton[] {
+        return [
+            {
+                type: 2,
+                style: 1,
```
Brainicism, line 1906: Can we make this a different color?

src/commands/game_options/timer.ts
taahamahdi: Since the timer is being reused, it should fall back to the default clip value when beyond this. Otherwise, clip mode would play very long clips of a song, making the feature redundant

src/commands/game_options/timer.ts
taahamahdi: This code branch is entered when not a game session, or a non-clip game mode is being used

src/enums/clip_action.ts
taahamahdi: This enum value is passed when the playSong is called for replays

src/commands/game_commands/play.ts
taahamahdi: It's user configurable during the game and only overwritten when it's either not set, or set to a value too large

src/structures/game_round.ts
taahamahdi: Yes, but it needs to be returned, otherwise it will exclude this button on the edit, so it can't be filtered out

src/structures/clip_game_round.ts
taahamahdi: I limited it here since it's not used in Round, but I can move it for future use

src/commands/game_options/timer.ts
taahamahdi: It can be simplified since `/timer` is scoped to only game sessions, thanks for finding it

src/structures/game_session.ts
taahamahdi: The `clipGameRound.getReplayCount() < 1` was an afterthought, I can take it out

src/constants.ts
taahamahdi: That's right. I planned on introducing an `/advanced` option to configure this so I preemptively moved it there

src/structures/session.ts
taahamahdi: > Are we assuming if its not CLIP mode, songStartedAt is non-null

Not quite. `songStartedAt` is used to track whether the round has been started, and this applies to both clip and non-clip modes. Behavior has not been changed for non-clip mode. However, for clip mode, it is only used to pick the initial starting location, but on a replay/new-clip, the other branches are entered.

To simplify this, I suggest we replace `!round.songStartedAt` with `!clipAction` -- this is the initial entry point to this function for both clip and non-clip game modes.

src/structures/session.ts
taahamahdi: Sorry, this was a remnant of a previous implementation

src/structures/session.ts
taahamahdi: I think this decision has led to weak points in this PR. However, I still think it makes the most sense for the end user. They are already familiar with changing the duration of a song with `/timer`.

If we move this value in as an argument to `/play`, the user loses the ability to customize this value during the game, and this will lead to shorter games with lost unique song history.

If we introduce another command scoped just for this, users may attempt to change its value using `/timer` and would need to learn another command. It would also mean we would need to track its state in the database, ignore showing it in `/options` except for during a clip game, and as a result, would not have an elegant way of changing its value outside a game.

src/structures/session.ts
taahamahdi: This also needs to be set when the round is starting, so it would need to be added to both of these branches:

```
        if (!round.songStartedAt) {
            switch (seekType) {
                case SeekType.BEGINNING:
                    seekLocation = 0;
                    break;
                case SeekType.MIDDLE:
                    seekLocation = songDuration * (0.4 + 0.2 * Math.random());
                    break;
                case SeekType.RANDOM:
                default:
                    seekLocation = songDuration * (0.6 * Math.random());
                    break;
            }
        } else if (clipAction === ClipAction.NEW_CLIP) {
            // Clip mode and the user requested another segment
            seekLocation = songDuration * (0.6 * Math.random());
```

I can move its initialization just below these if that makes it simpler

src/structures/session.ts
taahamahdi: I needed this to prevent casting `session` to a game session, as its currently scoped as `Session` here

src/structures/game_session.ts
taahamahdi: Sure, I can change it to be a secondary button (grey)

i18n/en.json
```diff
@@ -1475,6 +1486,7 @@
             "interaction": {
                 "alreadyEliminated": "You've already been eliminated this round.",
                 "bookmarkOutsideGame": "You can only bookmark songs during a game.",
+                "clipActionTooEarly": "Wait a bit before playing the clip.",
```
Brainicism, line 1489: Is this worded wrong?

i18n/en.json
```diff
@@ -1652,6 +1665,7 @@
             "title": "Wait..."
         },
         "releaseDate": "Release Date",
+        "replay": "Replay",
```
Brainicism, line 1668: Is this still used

src/constants.ts
```diff
@@ -166,6 +166,12 @@ export const ELIMINATION_DEFAULT_LIVES = 10;
 export const ELIMINATION_MAX_LIVES = 10000;
 export const ELIMINATION_MIN_LIVES = 1;
 
+export const CLIP_REPLAY_DELAY_MS = 100;
+export const CLIP_MIN_TIMER = 0.75;
```
Brainicism, line 170: _TIMER => LENGTH/DURATION?

src/commands/game_commands/play.ts
```diff
@@ -145,6 +158,29 @@ export default class PlayCommand implements BaseCommand {
...
+            {
+                example: clickableSlashCommand(COMMAND_NAME, GameType.CLIP),
+                explanation: i18n.translate(
+                    guildID,
+                    "command.play.help.example.clip",
+                    {
+                        clipTimer: `\`${CLIP_DEFAULT_TIMER}\``,
+                    },
+                ),
+            },
```
Brainicism, line 183: What's teh diff between these

src/commands/game_commands/play.ts
```diff
@@ -431,12 +519,16 @@ export default class PlayCommand implements BaseCommand {
...
+                ? null
+                : parsedMessage.components[1]!;
+
         await PlayCommand.startGame(
             MessageContext.fromMessage(message),
             gameType,
-            parsedMessage.components.length <= 1
-                ? null
-                : parsedMessage.components[1]!,
+            firstArg,
```
Brainicism, line 530: huh

src/structures/game_round.ts
```diff
@@ -362,50 +367,79 @@ export default class GameRound extends Round {
...
+            (actionRow) => ({
+                type: 1,
+                components: actionRow.components.map(
+                    (button: Eris.InteractionButton) => {
+                        if (
+                            Object.values(ClipAction).includes(
+                                button.custom_id as ClipAction,
+                            )
+                        ) {
+                            // TODO heart here
```
Brainicism, line 388: is this TODO still valid

src/structures/game_session.ts
```diff
@@ -266,7 +265,10 @@ export default class GameSession extends Session {
         await super.endRound(false, messageContext);
 
         try {
-            await round.interactionMarkAnswers(correctGuessers.length || 0);
+            await round.interactionMarkAnswers(
+                correctGuessers.length || 0,
```
Brainicism, line 269: Wouldnt .length return 0 anyway?

src/structures/game_session.ts
```diff
@@ -619,6 +621,56 @@ export default class GameSession extends Session {
...
+            i18n.translate(
+                this.guildID,
+                clipRound.isNewClipMajority()
+                    ? "misc.clip.newClip.success"
+                    : "misc.clip.newClip.request.description",
+                {
+                    voteCounter: clipRound.newClipVoteCounter(),
+                },
+            ),
+            !clipRound.isNewClipMajority(),
```
Brainicism, line 661: Why are we making this conditionally ephermeral? Are hints/skips ephermeral?

Brainicism (changes requested):

src/commands/game_commands/play.ts
```diff
@@ -1127,6 +1220,17 @@ export default class PlayCommand implements BaseCommand {
                 }
             }
 
+            let clipTimer: number;
```
Brainicism, line 1223: clip duration/length?

src/commands/game_commands/play.ts
```diff
@@ -1127,6 +1220,17 @@ export default class PlayCommand implements BaseCommand {
...
+            if (clipTimerArg == null) {
+                clipTimer = CLIP_DEFAULT_TIMER;
+            } else {
+                clipTimer = parseFloat(clipTimerArg);
+                clipTimer = Math.round(clipTimer! * 100) / 100;
+                if (clipTimer < CLIP_MIN_TIMER || clipTimer > CLIP_MAX_TIMER) {
+                    clipTimer = CLIP_DEFAULT_TIMER;
```
Brainicism, line 1230: Shouldn't it clip to the lower/upper bound instead of going to default?

src/structures/game_session.ts
```diff
@@ -140,6 +148,7 @@ export default class GameSession extends Session {
         this.songStats = {};
         this.lastGuesser = null;
         this.hiddenUpdateTimer = null;
+        this.clipTimerLength = clipTimerLength ?? 0;
```
Brainicism, line 151: Don't think a default value makes much sense here

src/commands/game_commands/play.ts
taahamahdi: This example showcases an argument being passed to `/play clip`

src/commands/game_commands/play.ts
taahamahdi: I cleaned it up a little, thanks

src/structures/game_round.ts
taahamahdi: I plan to get this resolved in https://github.com/Brainicism/KMQ_Discord/pull/2031

src/structures/game_session.ts
taahamahdi: They are not ephemeral, we can make these non-ephemeral as well if you'd prefer. Was hoping to cut down on chat spam in a short time interval

src/constants.ts
```diff
@@ -166,6 +166,14 @@ export const ELIMINATION_DEFAULT_LIVES = 10;
 export const ELIMINATION_MAX_LIVES = 10000;
 export const ELIMINATION_MIN_LIVES = 1;
 
+export const CLIP_MIN_DURATION = 0.75;
+export const CLIP_DEFAULT_DURATION = 2;
+export const CLIP_MAX_DURATION = 15;
```
Brainicism, line 171: Should we adjust these min/default/max values? They seem a bit high

src/structures/session.ts
```diff
@@ -745,15 +817,37 @@ export default abstract class Session {
...
+                if (clipAction !== ClipAction.END_ROUND) {
+                    logger.info(
+                        `${getDebugLogHeader(
+                            messageContext,
+                        )} | Song finished without being guessed.`,
+                    );
+                }
+            }
+
+            if (clipAction === ClipAction.END_ROUND) {
```
Brainicism, line 829: Can you add a comment on why we early returnh ere

src/structures/session.ts
```diff
@@ -769,10 +863,14 @@ export default abstract class Session {
                 this.connection.on("error", () => {});
             }
 
+            if (clipAction === ClipAction.END_ROUND) {
```
Brainicism, line 866: Same here

src/structures/game_session.ts
```diff
@@ -1652,16 +1759,32 @@ export default class GameSession extends Session {
...
+        if (this.isClipMode()) {
+            timestamp = Math.ceil(
+                (round.songStartedAt! +
+                    (CLIP_MAX_REPLAY_COUNT + 1) *
+                        (this.clipDurationLength! +
+                            CLIP_PADDING_BEGINNING +
+                            CLIP_PADDING_END +
+                            CLIP_VC_END_TIMEOUT_MS / 1000) *
+                        1000) /
+                    1000,
```
Brainicism, line 1772: extract this to a function, in session.ts since its used there too.

src/commands/game_commands/play.ts
```diff
@@ -1153,6 +1263,12 @@ export default class PlayCommand implements BaseCommand {
             }
         }
 
+        if (gameType === GameType.CLIP) {
+            if (guildPreference.isGuessTimeoutSet()) {
```
Brainicism, line 1267: Why do we set this to null? Is this used in clip mode?

Brainicism (changes requested):

src/commands/game_commands/play.ts
taahamahdi: Timer shouldn't be allowed with clip mode, so it's reset, and a precheck was added to prevent modifying /timer when in clip mode

Brainicism: If it's not allowed can't we just ignore it?

taahamahdi: It can be done by ignoring when the timer is fired. However, since this isn't a game option, I won't be able to mark it as a conflicting option, so it may cause confusion on why the timer isn't working when it's set

taahamahdi: I may be able to just hide the /timer entry in /options, let me try that

Brainicism: would try to avoid changing peoples options if possible

taahamahdi: Since there was already similar logic to ignore /goal in elimination, I followed that pattern and will show a conflict via game type

Brainicism: we should fix that eventually

Brainicism (approved):

PR #2038, opened 2024-03-31
Songs changed webhook prefer message over file

Brainicism:
File uploads are not previewable on mobile. Prefer embeds until length limit is reached
image

### DIFF
--- a/src/seed/seed_db.ts
+++ b/src/seed/seed_db.ts
@@ -3,7 +3,9 @@ import * as cp from "child_process";
 import {
     DATABASE_DOWNLOAD_DIR,
     DataFiles,
+    EMBED_DESCRIPTION_MAX_LENGTH,
     EMBED_ERROR_COLOR,
+    EMBED_SUCCESS_COLOR,
     KMQ_USER_AGENT,
     KmqImages,
 } from "../constants";
@@ -14,12 +16,13 @@ import {
     parseJsonFile,
     pathExists,
     standardDateFormat,
+    truncatedString,
 } from "../helpers/utils";
 import { getNewConnection } from "../database_context";
 import { program } from "commander";
 import {
     sendDebugAlertFileWebhook,
-    sendDebugAlertWebhook,
+    sendInfoWebhook,
 } from "../helpers/discord_utils";
 import { sql } from "kysely";
 import Axios from "axios";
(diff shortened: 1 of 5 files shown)


src/seed/seed_db.ts
```diff
@@ -734,22 +739,56 @@ async function reloadAutocompleteData(): Promise<void> {
             const currentDate = new Date();
             if (songsRemoved.length) {
                 logger.info(`${songsRemoved.length} songs removed.`);
-                await sendDebugAlertFileWebhook(
-                    discordDateFormat(currentDate, "f"),
+                const description = `**${songsRemoved.length} songs removed**:\n${songsRemoved.join("\n")}`;
```
taahamahdi, line 742: Do we want to apply some formatting to make it more readable

src/seed/seed_db.ts
```diff
@@ -734,22 +739,56 @@ async function reloadAutocompleteData(): Promise<void> {
...
+                    truncatedString(
+                        description,
+                        EMBED_DESCRIPTION_MAX_LENGTH - 1,
+                    ),
+                    EMBED_ERROR_COLOR,
+                    KmqImages.NOT_IMPRESSED,
+                    undefined,
                 );
+
+                if (description.length >= EMBED_DESCRIPTION_MAX_LENGTH) {
```
taahamahdi, line 756: Should we fall back to only the file upload when truncated instead of duplicating data?

src/seed/seed_db.ts
Brainicism: What more formatting is needed? Also more formatting => less entries

src/seed/seed_db.ts
Brainicism: Is this not what its doing?

src/seed/seed_db.ts
taahamahdi: It gets a bit difficult to read on mobile. Maybe we can put them in a bulleted list

src/seed/seed_db.ts
taahamahdi: I believe this is sending an embed with the truncated songs, and then sending the full list as a file. We could send only the file if it's too long, similar to how ,list works

Brainicism: Would rather still show some/most of the list in the original message, since mobile users cant see file previews at all

taahamahdi: Good point, let's leave the current implementation

taahamahdi (approved):

PR #2105, opened 2024-04-22
Add /bookmark to see previous bookmarks

taahamahdi:
Screenshot 2024-04-21 at 10 04 54 PM
Screenshot 2024-04-21 at 10 05 02 PM
Should we filter out duplicates and only show the latest bookmark?


src/commands/misc_commands/bookmark.ts
```diff
@@ -0,0 +1,354 @@
...
+import type BaseCommand from "../interfaces/base_command";
+import type CommandArgs from "../../interfaces/command_args";
+import type HelpDocumentation from "../../interfaces/help";
+import type MatchedArtist from "src/interfaces/matched_artist";
+
+const COMMAND_NAME = "bookmark";
+const logger = new IPCLogger(COMMAND_NAME);
+
+// eslint-disable-next-line import/no-unused-modules
+export default class BookmarkCommand implements BaseCommand {
```
Brainicism, line 40: 'bookmark' sounds like an action to create a bookmark, not an action to list bookmarks. Should change every reference of 'bookmark' to 'bookmarks'

Brainicism (changes requested):

Brainicism (approved):

PR #2129, opened 2024-05-01
Remove dependency on servers/channels on different clusters

Brainicism:
- Replace power hour/debug server messages with webhook calls
- Replace debug server permissions with admin permissions


src/helpers/discord_utils.ts
```diff
@@ -1698,6 +1718,31 @@ export async function sendInfoWebhook(
...
+ */
+export async function sendInfoEmbedsWebhook(
+    webhookURL: string,
+    embed: EmbedPayload,
+    content: string | undefined,
+): Promise<void> {
+    if (!webhookURL) return;
+    await axios.post(webhookURL, {
+        content,
+        embeds: [
```
taahamahdi, line 1735: Does this need a thumbnail URL? Is it missing any default fields send info message has? Can you use the spread operator to fill embed fields instead?

Brainicism: Doesn't _need_ it, will just default to the webhook's default avatar/username, but added it as a param anyway. Can't usre spread operator because the field names don't match exactly

taahamahdi (approved):

taahamahdi (approved):

PR #2174, opened 2024-08-02
Add error handler on readable stream

Brainicism:


src/scripts/download-new-songs.ts
```diff
@@ -165,9 +166,18 @@ const downloadSong = (db: DatabaseContext, id: string): Promise<void> => {
...
-                cacheStream,
+            const ytdlReadableStream = ytdl(
+                `https://www.youtube.com/watch?v=${id}`,
+                ytdlOptions,
             );
+
+            ytdlReadableStream.on("error", (err: Error) => {
+                const errorMessage = `Error in ytdl readable stream. err = ${extractErrorString(err)}`;
+                logger.error(errorMessage);
+                reject(new Error(errorMessage));
```
taahamahdi, line 177: Should return after this reject

Brainicism: It wouldn't really do anything because we're inside of a callback, the event loop is already past this line of code.

taahamahdi (approved):

PR #2232, opened 2024-09-26
Fix accepted artist name answers for artist with + in name

Brainicism:
`song.artistName.split("+")` was returning an empty string for `24K+` and `Mamamoo+`. Check for collabs using `% + %` instead.

### DIFF
--- a/src/structures/game_round.ts
+++ b/src/structures/game_round.ts
@@ -267,7 +267,6 @@ export default class GameRound extends Round {
         guessModeType: GuessModeType,
         typosAllowed = false,
     ): number {
-        // if (!guess) return 0;
         let pointReward = 0;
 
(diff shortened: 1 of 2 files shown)


src/structures/game_round.ts
```diff
@@ -262,6 +267,7 @@ export default class GameRound extends Round {
         guessModeType: GuessModeType,
         typosAllowed = false,
     ): number {
+        // if (!guess) return 0;
```
taahamahdi, line 270:
```suggestion
```

taahamahdi (dismissed):

taahamahdi (approved):

taahamahdi (approved):

taahamahdi (approved):

PR #2276, opened 2025-06-05
Add check for completely empty better audio column

Brainicism:
There's a strange bug where the `better_audio_column` is seemingly empty, despite Daisuki data looking good. This causes the seed to delete all of those videos that previously had a better audio, since it thinks it never had one and needs to download a new one.


taahamahdi (approved):

taahamahdi (approved):
LGTM. We should follow-up with a change that replaces every `rm` with `mv x /tmp` to ensure we don't have to redownload in the future

PR #2326, opened 2026-04-30
Fix 6 lower-severity bugs: majority count, raw JSON display, UUID correlation, timer drift, scoreboard layout, weakest player crash

Brainicism:
## Summary

Six lower-severity bug fixes across helpers, structures, and scoreboard logic.

---

### 1. `getMajorityCount` returns 0 during voice disconnects
**File:** `src/helpers/discord_utils.ts`
**Bug:** When the bot's voice connection is momentarily unavailable, `getMajorityCount` returns `0`. Since `skipCount >= 0` is always true, a single skip vote instantly skips the round.
**Impact:** During brief VC reconnections, any single skip vote immediately skips the round, bypassing the majority requirement.
**Fix:** Return `Number.MAX_SAFE_INTEGER` instead of `0` when the voice connection is unavailable, so no single vote can trigger a skip.

---

### 2. `getGameInfoMessage` returns raw JSON to users on parse failure
**File:** `src/helpers/discord_utils.ts`
**Bug:** When `JSON.parse` succeeds but locale lookup or other processing fails in the catch block, the original raw JSON string (e.g., `{"en":"Hello","ko":"안녕"}`) remains in the message/title fields and is returned to the caller.
**Impact:** Users occasionally see raw JSON objects instead of readable text in game info messages.
**Fix:** Return `null` in both catch blocks so broken messages are suppressed rather than shown as raw JSON.

---

### 3. `retryWithExponentialBackoff` generates different UUIDs per log line
**File:** `src/helpers/utils.ts`
**Bug:** Each `uuid.v4()` call inside the function generates a new UUID. The initial log, retry logs, and failure log all have different UUIDs for the same operation.
**Impact:** When debugging retry sequences in logs, it's impossible to correlate which retries belong to the same original operation.
**Fix:** Generate the UUID once at the top of the function (`const operationId = uuid.v4()`) and reuse it across all log lines.

---

### 4. `timerStartedAt` set at Round construction, not when playback starts
**File:** `src/structures/session.ts`
**Bug:** `timerStartedAt` is set to `Date.now()` in the Round constructor, but the actual timer doesn't begin until after voice connection setup and song loading (where `songStartedAt` is set).
**Impact:** Hidden-mode countdown displays expire visually before the actual timeout fires, showing inaccurate remaining time to players.
**Fix:** Set `round.timerStartedAt = Date.now()` alongside `round.songStartedAt` when playback actually begins.

---

### 5. Three-column scoreboard first column not inline
**File:** `src/structures/scoreboard.ts`
**Bug:** `getScoreboardEmbedThreeColumns` has `inline: false` on the first embed field, while the second and third fields have `inline: true`.
**Impact:** The scoreboard renders as a full-width first row followed by two side-by-side columns, instead of the intended three-column layout.
**Fix:** Changed first field to `inline: true` to match the other two columns.

---

### 6. `getLivesOfWeakestPlayer` crashes when all players are eliminated
**File:** `src/structures/elimination_scoreboard.ts`
**Bug:** `.filter((x) => x.getLives() > 0)` produces an empty array when all players are dead, then `.reduce()` on an empty array with no initial value throws `TypeError: Reduce of empty array with no initial value`.
**Impact:** Game crashes with TypeError when checking weakest player after all players have been eliminated.
**Fix:** Added early return of `0` if no alive players exist before calling `.reduce()`.

---

### 7. `getDurationFooter` treats `timeRemaining === 0` as no duration
**File:** `src/structures/session.ts`
**Bug:** `if (!timeRemaining)` evaluates to `true` when `timeRemaining` is exactly `0` (falsy), causing the function to return an empty string.
**Impact:** The "⏰ Time finished!" message never displays when the timer is exactly at 0.
**Fix:** Changed to `if (timeRemaining == null)` so `0` is treated as a valid value.


src/helpers/discord_utils.ts
```diff
@@ -1684,7 +1690,7 @@ export function getMajorityCount(guildID: string): number {
         return Math.floor(getNumParticipants(voiceChannelID) * 0.5) + 1;
     }
 
-    return 0;
+    return Number.MAX_SAFE_INTEGER;
```
taahamahdi: If the VC is bugged and players attempt to skip, will this make it impossible to go to the next song?

src/structures/scoreboard.ts
```diff
@@ -335,7 +335,7 @@ export default class Scoreboard {
                 value: players
                     .slice(0, Math.ceil(players.length / 3))
                     .join("\n"),
-                inline: false,
+                inline: true,
```
taahamahdi: This is by design. It looks weird when putting them all inline

src/structures/session.ts
```diff
@@ -1210,7 +1211,7 @@ export default abstract class Session {
         timeRemaining: number | null,
         nonEmptyFooter: boolean,
     ): string {
-        if (!timeRemaining) {
+        if (timeRemaining == null) {
```
taahamahdi: This is modified by  #2325

Brainicism (commented):
Addressed all 3 comments:

1. **getMajorityCount** — Good call, changed from `MAX_SAFE_INTEGER` to `1`. Still prevents the 0-threshold instant-skip bug, but allows a single deliberate skip vote to go through when VC is bugged.

2. **Scoreboard inline** — Reverted. Intentional design choice, not a bug.

3. **getDurationFooter** — Removed, already handled by #2325.

PR #2329, opened 2026-04-30
Add Discord Activity (Embedded App) integration

taahamahdi:
In-Discord iframe Activity that runs alongside the bot's voice playback, delivering live scoreboard, click-to-guess, hint/skip/bookmark voting with progress bars, and game-lifecycle controls (start/end). The bot remains authoritative for game state; the Activity is a presentation and input layer wired through a new admiral-side hub.

Server:
- ActivityHub on the admiral with per-guild WebSocket subscribers and a correlation-ID promise map for request/reply IPC routed to the worker cluster owning each guild.
- activity_bridge on workers instruments Session lifecycle events (roundStart, roundEnd, scoreboardUpdate, guessReceived, sessionEnd, hintProgress, hintRevealed, skipProgress, skipped) and dispatches admiral requests (snapshot, guess, startGame, skipVote, endGame, hint, bookmark) by reusing existing command methods with a synthetic MessageContext — no command refactor needed.
- /api/activity/{token,session,start,skip,end,hint,bookmark,guess} with Discord OAuth2 token exchange, activity-instance participant verification (handles both string-snowflake and {id} response shapes), and a /ws/activity push channel with 30s heartbeat. Static handler serves activity/dist; GET / serves index.html so Discord's iframe entry resolves.

Frontend (new activity/ workspace, Vite + React + @discord/embedded- app-sdk):
- Authenticates via DiscordSDK, opens a WS, applies an initial snapshot, and reduces live events into UI state.
- Live scoreboard with rank changes, current-round timer, guess ticker with usernames resolved from the scoreboard, song reveal with thumbnail proxied through Discord's URL Mapping, all-guesses panel showing each player's last guess, hint/skip/bookmark vote tally bars, and a responsive two-column round area that reserves space so the layout doesn't shift between guess and reveal.
- Auto-focus the guess input on round transitions and after submit.
- Thumbnail click opens YouTube via openExternalLink (iframe CSP).

Tooling:
- npm run build:activity (yarn install + vite build).
- npm run tunnel: cloudflared helper that captures the URL and updates ACTIVITY_PUBLIC_BASE_URL in .env.
- npm run repair-available-songs: schema repair script for when the available_songs table drifts from the stored procedures.

Also fixes a pre-existing circular-JSON crash in Session.startRound's error logging that was masking the underlying DB error.


Brainicism (commented):
## Review: Discord Activity Integration

Big feature — architecture is solid. Admiral↔worker IPC with correlation-ID promises is clean, WebSocket heartbeat handling is good, and "bot authoritative, Activity as presentation layer" is the right call. A few things to address before merge and some follow-ups.

---

### 🔴 Fix Before Merge

**SEC-1: OAuth token leaked in WebSocket URL** (`activity/src/api.ts` — `openActivityStream`)

The Discord access token is passed as `?access_token=...` in the WS query string. This exposes it in server access logs, browser dev tools, proxy/CDN logs, and potentially browser history. Consider a ticket exchange: hit a REST endpoint to swap the access token for a short-lived opaque ticket (random UUID, valid ~10s, stored server-side), then pass only the ticket in the WS URL.

**STATE-1: Snapshot `guessTimeoutSec` always `null`** (`activity_bridge.ts` — `buildSessionSnapshot`)

```ts
guessTimeoutSec: null,  // ← hardcoded
```

Late joiners and reconnecting clients never see the countdown timer because the snapshot hardcodes `null`. Should read from `session.guildPreference` the same way the `roundStart` emit does.

**STATE-3: `sessionStart` doesn't reset `bookmarkedLinks`** (`activity/src/App.tsx` — `reduce()`)

Starting a new game carries over stale bookmarks from the previous session. Add `bookmarkedLinks: new Set(), roundBookmark: { pending: false, bookmarked: false }` to the `sessionStart` reducer.

---

### 🟠 Should Fix

**RACE-1: Bridge commands don't participate in lifecycle mutex** (`activity_bridge.ts`)

All mutating bridge operations are fire-and-forget — they reply `ok: true` before the async work completes. Two simultaneous `startGame` requests from the Activity can both pass the session-exists check. At minimum, `startGame` should acquire a per-guild lock. (See #2328 which adds lifecycle mutexes.)

**RACE-2: `attachActivityBridge(this)` in constructor** (`game_session.ts`)

This pushes a `sessionStart` event before the session is registered in `State.gameSessions`. If a WS subscriber reacts to the event and queries the session registry, it won't find it. Move to after registration, or lazy-attach on first `startRound`.

**RACE-3/4: Skip and hint handlers read `session.round` after async gap** (`activity_bridge.ts`)

After `executeSkip()`/hint resolves, `session.round` might be the *next* round if a transition happened during the await. Capture a reference before the async call and compare by identity:
```ts
const originalRound = session.round;
await SkipCommand.executeSkip(messageContext);
if (session.round !== originalRound) { /* round transitioned */ }
```

**SEC-2/3: No rate limiting on activity endpoints** (`kmq_web_server.ts`)

The token exchange and all action endpoints (`/api/activity/*`) are wide open. An attacker spamming these can exhaust Discord API rate limits on the bot's credentials. Even a generous `fastify-rate-limit` (e.g., 30/min/IP on token, 60/min/IP on actions) would prevent abuse.

---

### 🟡 Follow-ups / Non-blocking

- **IPC-1:** `accessTokenCache` and `instanceCache` grow unboundedly — add periodic cleanup or use an LRU with max size
- **WS-1:** No reconnect logic in frontend — transient network blip = manual refresh. Exponential backoff reconnect + snapshot re-fetch would be a nice phase 2
- **WS-2:** EventEmitter listeners on Session never explicitly cleaned up on session end (GC handles it, but explicit `removeAllListeners()` is cleaner)
- **STATE-2:** `sessionEnd` reducer doesn't clear `session` — stale metadata lingers
- **CQ-1:** Import cycle `game_session ↔ activity_bridge` (eslint-disabled, runtime-safe but fragile with ESM)
- **CQ-2:** Duplicate type definitions between server (`activity_types.ts`) and client (`types.ts`) — will drift over time
- **CQ-3:** `any` casts on request body/query instead of Fastify schema validation
- **CQ-4:** `repair-available-songs.ts` passes DB password as CLI arg (visible in `ps`) — use `--defaults-file` or `MYSQL_PWD` env var
- **NIT:** `RoundTimer` re-renders every 250ms but only displays whole seconds — 1000ms interval would cut 75% of no-op renders

---

Overall: clean architecture, well-structured frontend, correct Discord SDK integration. Main concerns are the token-in-URL leak and the race conditions that should participate in the lifecycle mutex from #2328. 👍

src/scripts/repair-available-songs.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable no-console */
```
Brainicism: Do we need to commit this file?

taahamahdi: It's not relevant to this change at all. I just removed it

src/commands/game_commands/play.ts
```diff
@@ -41,6 +41,7 @@ import AnswerCommand from "../game_options/answer";
 import AnswerType from "../../enums/option_types/answer_type";
 import CommandPrechecks from "../../command_prechecks";
 import Eris from "eris";
+// eslint-disable-next-line import/no-cycle
```
Brainicism: We probably shouldn't do this.. we spent days removing all of the cycles last time which were causing all sort of race conditions

taahamahdi: Oh right. I'll clean it

Brainicism (approved):

Brainicism:
# Code Review — High-Priority Findings

8 high-priority findings from a two-pass deep review of this PR. Grouped by urgency.

---

## 🔴 Must-Fix

### H-2. Unbounded cache growth — memory leak

`accessTokenCache`, `instanceCache`, `wsTicketCache` in `kmq_web_server.ts` are plain `Map`s with TTL-on-read only. There's no periodic sweep and no size cap. A long-running bot process will accumulate stale entries indefinitely → eventual OOM.

**Recommendation:** Add a periodic `setInterval` sweep that evicts expired entries, and cap max entries with LRU eviction.

---

### H-3. `session.round` already null in `roundEnd` handler

In `game_session.ts`, `super.endRound()` nulls `this.round` before the `roundEnd` event fires. The bridge's `roundEnd` handler reads `session.round.songStartedAt` → always `null` → `allGuesses[].ts` values are relative ms instead of absolute timestamps.

```ts
// game_session.ts — current flow:
super.endRound();        // ← nulls this.round
this.emit("roundEnd", {  // ← handler tries to read session.round.songStartedAt → null
    songInfo: ...,
    allGuesses: ...,
});
```

**Recommendation:** Capture `round` data before calling `super.endRound()`, or emit the event before the super call.

---

### H-5. Dead WS subscribers never removed

In `kmq_web_server.ts`, when `ws.send()` fails on a subscriber, it logs a warning but doesn't remove the socket from the subscriber `Set`. Dead sockets accumulate → warning log spam on every event broadcast indefinitely.

**Recommendation:** On send failure, remove the socket from the subscriber set immediately:

```ts
try {
    ws.send(JSON.stringify(event));
} catch (e) {
    logger.warn("Failed to send WS event", e);
    subscribers.delete(ws);  // ← add this
}
```

---

### H-8. OAuth token never refreshed — silent auth failure after expiry

In `activity/src/App.tsx` and `activity/src/discordSdk.ts`, the OAuth access token is obtained once during `authenticate()` and stored in React state. No refresh/retry logic exists. When the token expires or is revoked, all API calls silently return 401 with no way for the user to re-authenticate without closing and reopening the Activity.

**Recommendation:** Add a 401 interceptor that re-authenticates via the Discord SDK and retries the failed request. At minimum, show a "session expired, please reload" message instead of silently failing.

---

## 🟠 Should-Fix

### H-1. No WebSocket reconnection — connection loss is terminal

In `activity/src/App.tsx`, if the WS connection drops (network blip, server restart), there's no reconnection logic. The Activity becomes a dead UI. Users must close and reopen — unacceptable for 10-30 min game sessions.

**Recommendation:** Add exponential backoff reconnection with snapshot re-fetch on reconnect.

---

### H-7. Cached participant list locks out late joiners

In `kmq_web_server.ts`, `resolveActivityInstance()` caches the Discord API response including `participantIDs` for 5 seconds. A user who joins the VC after the cache is populated will authenticate successfully (separate OAuth), load the snapshot, and see the game running — but every action (guess, skip, hint) silently returns 403 until the cache expires.

**Recommendation:** Either remove participant validation entirely (the VC membership check on the worker side already prevents non-VC users from acting), or invalidate the cache entry when a 403 would be returned so the next request gets a fresh participant list.

---

### H-4. `guess` operation excluded from guild lock — verify concurrency safety

In `activity_bridge.ts`, guess operations bypass `withGuildLock()` intentionally for latency. The existing `lifecycleMutex` likely makes this safe, but needs explicit verification that concurrent guesses can't corrupt `GameRound` state (e.g., `correctGuessers` array, score updates).

**Recommendation:** Add a code comment documenting why this is safe, or add a lightweight per-round lock for guess processing.

---

### H-6. Stale `t` closure in WS `onClose` callback

In `kmq_web_server.ts`, the `t` (translator) function is captured in a closure when the WS connection is established. If the i18n bundle loads after mount and changes the locale resolution, the disconnect message displays in the wrong language.

**Recommendation:** Resolve `t` at call time instead of capturing it in the closure.

taahamahdi:
Following up with the rest on the #2342

PR #2342, opened 2026-05-02
Activity Phase 4 + Phase 5: full options panel, UI redesign, reduced channel embeds

taahamahdi:
Builds on #2329 (Discord Activity integration). Base branch: `master`.

Completes Phase 4 (configure all high-traffic `GuildPreference` settings from the Activity) and lands Phase 5 (reduce channel embeds when users move to the Activity), plus a UI redesign, song history, dark mode, and assorted infrastructure cleanup.

## Phase 4 — full options panel

All four slices land here; each setter still works unchanged through slash commands, and every write — slash or Activity — broadcasts an `optionsChanged` wire event to all open Activity tabs in the guild via a module-level EventEmitter (`src/helpers/guild_preference_events.ts`).

### Slice 1 — flat enums
- `gender` (chip toggles + mutually exclusive `alternating`), `guessMode` (radio), `multiguess` (toggle).

### Slice 2 — numeric options
- `limit` (range 0–100 000, start < end), `cutoff` (year range 1900–current), `goal` (1–100 000 or unset), `timer` (2–180s or off).

### Slices 3 & 4 — autocomplete + duration
- `groups`, `include`, `exclude` — autocomplete-driven artist selection. Autocomplete is routed admiral → cluster-0 worker via the existing ActivityHub IPC pattern (`src/activity_hub.ts:autocompleteArtists`); admiral doesn't duplicate artist state.
- `duration` — add/set/remove semantics matching the slash command.

### Validation & safety
- `POST /api/activity/option` whitelists `kind` + value shape + bounds per option. Malicious clients can't persist out-of-range values.
- The `setOption` worker handler runs under the existing per-guild lifecycle lock and rejects on maintenance / ban / not-in-VC (session-only) before calling the matching `GuildPreference` setter.
- Snapshot now carries `options`, loaded upfront so lobby guilds (no active session) see current values.

## Phase 5 — reduced channel embeds

Gated behind a new `KmqConfiguration.activityReducedEmbeds()` flag (default off; flip via `data/feature_switch_config.json` + `reload_config` IPC). When on:

- Round reveal embed (`GameSession.endRound`): collapses to a one-line "open the Activity" pointer with song / artist / winner inline.
- End-of-game summary (`GameSession.sendEndGameMessage`): collapses to one line with winner / footer.
- Skip vote tallies (`SkipCommand.sendSkipNotification`): suppressed (the "skipped!" confirmation still fires for non-Activity users).
- Hint vote tallies (`HintCommand.sendHintNotification`): suppressed (the hint reveal itself still fires).

Caveat (intentional, documented): the flag is per-guild, not per-user. Text channels are shared, so once anyone in the guild opens the Activity, the channel goes minimal for everyone. Recommendation is to leave the flag off for mixed-mode guilds; the infrastructure is there for guilds that have collectively moved to the Activity.

## UI redesign

- **Cream / pink palette** with two collapsible sidebars (options on the right, song history on the left, both open by default).
- **Song-history left sidebar** — running log of past rounds inside the Activity.
- **Dark mode toggle** (☀/☾) next to the sidebar toggles. Choice persisted in localStorage, defaults to dark on first launch. Palette split into `:root[data-theme=\"light\"]` / `[data-theme=\"dark\"]` reading the same custom-property names.
- **Guess input clears on round start** — previously kept the stale text from the last round.
- **SDK reject panel** reflowed to one line.

### Docker
- Dockerfile gains an activity-build step between `tsc` and runtime: copies `activity/` sources, runs `yarn install --frozen-lockfile && yarn build`, emits `activity/dist`.
- Vite `envPrefix` widened to `[\"VITE_\", \"BOT_CLIENT_ID\"]` so the bundle reads `import.meta.env.BOT_CLIENT_ID` directly — same env var the server reads at runtime. Operators set `BOT_CLIENT_ID` once.
- `docker-compose.yml` forwards `BOT_CLIENT_ID` as a build ARG; `docker/.env.sample` documents `BOT_CLIENT_ID` / `DISCORD_CLIENT_SECRET` / `ACTIVITY_PUBLIC_BASE_URL` plus which file flows into which build path.

### YouTube thumbnail fallback
- Activity reveal probes the YouTube thumbnail chain (`maxresdefault → sddefault → hqdefault → mqdefault → default`) via off-DOM `Image()` so the visible `` only mounts once a real image loads. Catches both real 404s and the case where the proxy normalizes 404 → 200 (YouTube serves a 120×90 \"no preview\" placeholder body, which the natural-width check rejects).

### i18n tooling
- Bumped `i18n-ai-translate` to v5.0.1 (the v4 `diff` subcommand had two bugs: data loss on merge + BCP-47 rejection; both fixed upstream).
- Backfilled `activity.options.*` and dark-mode tooltip strings into all 11 non-English locales via `i18n-ai-translate diff`.
- Cleaned up the longstanding `missing_i18n` dynamic-key warning in `GameSession.sendActivityReducedRoundReveal` by inlining the conditional so the linter walks both literals directly.

## Verification

Manual end-to-end (Docker image + local tunnel):

1. Activity iframe loads against a tunneled HTTPS endpoint.
2. Two Activity tabs in the same guild: any setter toggled in tab A propagates to tab B within ~200ms. Slash-command writes (`/gender male`) propagate to both tabs.
3. All seven Phase-4 options edit + persist; invalid payloads (range violations, unknown enum values) → 400, no DB write.
4. `groups` / `include` / `exclude` autocomplete returns results from cluster 0 via IPC.
5. Phase-5 flag on: round / end / skip-vote / hint-vote channel embeds collapse to single-line pointers. Slash-command-driven `/score` and multiple-choice button messages stay untouched.
6. Dark mode persists across reloads; sun/moon icons swap; defaults to dark on first launch.
7. Thumbnail fallback walks past missing `maxresdefault` and proxy-served placeholders without flashing the placeholder.

CI expected to pass on `lint-ci`, `prettier-ci`, server `tsc`, activity `tsc` + `vite build`, `lint-i18n-ci`, and Docker build.

## Out of scope (later phases)

- **Lower-traffic options** — language, OST, special, shuffle, artist type, answer type, release type, seek, subunit. Slash-only for now.
- **Per-user reduced-embeds gating** — fundamentally impossible (text channels are shared); flag stays per-guild.
- **Synced YouTube iframe** at round end (replace static thumbnail with muted iframe at VC timestamp) — prototyped feasibility, not implemented.

## Test plan

- [ ] `docker-compose-dev` boots with `BOT_CLIENT_ID` / `DISCORD_CLIENT_SECRET` set.
- [ ] Activity iframe loads against tunneled HTTPS.
- [ ] All seven enum/numeric options edit; autocomplete-backed group options edit.
- [ ] Slash-command writes propagate to open Activity tabs.
- [ ] Invalid payloads → 400.
- [ ] Phase-5 flag on / off swaps embed verbosity at runtime via `reload_config`.
- [ ] Dark mode toggle persists across activity reloads.
- [ ] Existing slash commands unchanged.


Brainicism:
## Code Review — Major Issues Only

Solid PR overall. Input validation is thorough (whitelist enums, bounds checks, 200-artist cap, server-injected guildID/userID), the voice channel security fix properly reads user voice state instead of trusting the bot's member cache, and client-side optimistic updates with rollback are clean.

Two things worth noting:

### ⚠️ Reduced embeds is a global kill switch

`activityReducedEmbeds()` in `hint.ts` / `skip.ts` suppresses embeds for **all** guilds, not just ones with active Activity sessions. Text-only guilds lose their hint/skip notification messages when the flag is on. Should probably be gated per-session (e.g. `session.hasActivitySubscribers`) rather than server-wide config. Not a bug since it defaults to `false`, but could bite if rolled out broadly.

### ⚠️ `resolveArtistIDs` rebuilds the full Map every call

Iterates the entire `State.artistToEntry` and creates a new `Map<number, MatchedArtist>` on every `setOption` call for groups/includes/excludes. Under the lifecycle lock so it's safe, but wasteful with a large artist cache. Could cache the `byID` map alongside `State.artistToEntry` and invalidate on `reloadCaches`.

### ℹ️ Minor notes (non-blocking)

- `cutoff` fires two sequential DB writes → two `optionsChanged` events. Reducer is idempotent so it's fine, but users might see a flash of intermediate state.
- `duration` cast (`as number`) masks the `number | null` union — worth confirming `setDuration(null)` actually works as a reset.
- Autocomplete always routes to cluster 0 — if that worker is down, all autocomplete silently returns empty. Round-robin might be more robust.

Brainicism (approved):

taahamahdi:
Thanks for the review. Pushed `2ecde455` addressing two of these:

**`duration` cast (`as number`)** — good catch, and it was hiding more than a union. `setDuration` was already being handed `null` by the reset machinery (`resetArgs[DURATION].default = [null]`), so the `number`-only signature was a lie regardless of the Activity. Widened it to `number | null` (matching `setGoal`/`setGuessTimeout`) and dropped the cast — the Activity's "clear" now goes through the exact same path `/duration reset` uses.

**`resolveArtistIDs` rebuilding the Map** — done as suggested. Added `State.artistIDToEntry`, rebuilt in `reloadArtists` (cleared at the top so removed artists don't linger), and `resolveArtistIDs` just reads it now. No more full-cache iteration per `setOption`.

**Reduced embeds as a global kill switch** — agree this is the one to watch, but I left it for a follow-up rather than rush it here. The proper fix (gate on something like `session.hasActivitySubscribers`) is real cross-process plumbing: suppression runs on the worker, but subscriber state only exists in `ActivityHub` on the admiral, so it needs subscribe/unsubscribe deltas pushed admiral→worker plus the reconnect/restart/multi-tab races handled — its own change with tests. Since it defaults to `false` it's not load-bearing yet; I'd rather not ramp it broadly until that gating exists.

On the remaining minor notes: the `cutoff` double-write is idempotent so I left it; autocomplete cluster-0 routing is a broader robustness change I'd rather not fold into this PR.

PR #2345, opened 2026-05-02
Fix re-entrant lifecycle mutex deadlock (regression from #2328)

Brainicism:
### Root Cause

The `lifecycleMutex` added in #2328 wraps `startRound()`, `endRound()`, and `endSession()` in `GameSession`. However, `async-mutex` Mutex is **not re-entrant** — nested `runExclusive()` calls on the same mutex wait forever.

Multiple code paths inside `Session` (the parent class) call `this.endSession()` from within lifecycle methods that are already holding the mutex:

| Outer holder | Inner call | Trigger |
|---|---|---|
| `startRoundCore` → `super.startRound()` | `this.endSession()` | Maintenance mode, song reload error, null song, empty VC, VC connection failure (5 paths) |
| `endRoundCore` → `super.endRound()` | `this.endSession()` | Game duration reached |
| `startRoundCore` → `super.startRound()` → `playSong()` catch | `this.errorRestartRound()` → `this.endRound()` + `this.startRound()` | Audio playback error |

Since `this` is a `GameSession`, these polymorphically dispatch to the mutex-wrapped public methods — deadlock.

### Impact

A deadlocked session becomes a permanent zombie:
- `this.finished` never set to `true`
- `Session.deleteSession()` never called — session stays in `State.gameSessions`
- `cleanupInactiveGameSessions` tries `endSession()` every 10 min but can never acquire the mutex
- Session persists indefinitely (observed: 1686 minutes / 28 hours)

### Fix

**Lifecycle hooks:** Added three protected methods to `Session` (`endSessionFromLifecycle`, `endRoundFromLifecycle`, `startRoundFromLifecycle`) that default to calling the public API. `GameSession` overrides them to call `*Core` methods directly, bypassing the mutex since the caller already holds it.

All `this.endSession()`/`this.endRound()`/`this.startRound()` calls within `Session.startRound()`, `Session.endRound()`, and `Session.errorRestartRound()` now route through these hooks.

**Deadlock detection:** All three public mutex-wrapped methods in `GameSession` now log `POTENTIAL DEADLOCK` at error level if the mutex is already held when called.

**Cleanup safety net:** `cleanupInactiveGameSessions` now races `endSession()` against a 30s timeout. If `endSession` doesn't complete in time, the session is force-removed from `State.gameSessions`.

### Exhaustive audit — no remaining deadlock paths

**Paths fixed (previously deadlocked):**
- `Session.startRound()` → `this.endSessionFromLifecycle()` × 5 error paths ✅
- `Session.endRound()` → `this.endSessionFromLifecycle()` × 1 path ✅
- `Session.errorRestartRound()` → `this.endRoundFromLifecycle()` + `this.startRoundFromLifecycle()` ✅

**Paths already safe (use `*Core` directly, no mutex re-entry):**
- `endRoundCore` → `this.endSessionCore()` (game finished / sudden death) ✅
- `endRoundCore` → `this.startRoundCore()` (next round) ✅
- `endSessionCore` → `super.endSession()` (parent method, no mutex) ✅

**Paths outside mutex (acquire normally, no re-entry):**
- `guessTimeoutFunc` setTimeout → `this.endRound()` + `this.startRound()` ✅
- `connection.once("end")` → `this.endRound()` + `this.startRound()` ✅
- `connection.once("error")` → `this.errorRestartRound()` ✅
- `guessSong()` → `this.endRound()` + `this.startRound()` ✅
- `cleanupInactiveGameSessions` → `gameSession.endSession()` ✅

**ListeningSession:** Does not use `lifecycleMutex`, not affected. ✅

### DIFF
--- a/src/structures/game_session.ts
+++ b/src/structures/game_session.ts
@@ -262,7 +262,7 @@ export default class GameSession extends Session {
     ): Promise<void> {
         const waitStart = Date.now();
         try {
-            return await this.lifecycleMutex.runExclusive(() => {
+            await this.lifecycleMutex.runExclusive(async () => {
                 const waitMs = Date.now() - waitStart;
                 if (waitMs > 5000) {
                     logger.warn(
@@ -270,7 +270,7 @@ export default class GameSession extends Session {
                     );
                 }
 
-                return this.endRoundCore(isError, messageContext, gameRound);
+                await this.endRoundCore(isError, messageContext, gameRound);
             });
         } catch (e) {
             if (e === E_TIMEOUT) {
@@ -295,7 +295,7 @@ export default class GameSession extends Session {
     async endSession(reason: string, endedDueToError: boolean): Promise<void> {
         const waitStart = Date.now();
         try {
-            return await this.lifecycleMutex.runExclusive(() => {
+            await this.lifecycleMutex.runExclusive(async () => {
                 const waitMs = Date.now() - waitStart;
                 if (waitMs > 5000) {
                     logger.warn(
@@ -303,7 +303,7 @@ export default class GameSession extends Session {
                     );
                 }
 
-                return this.endSessionCore(reason, endedDueToError);
+                await this.endSessionCore(reason, endedDueToError);
             });
         } catch (e) {
             if (e === E_TIMEOUT) {
--- a/src/structures/game_session.ts
+++ b/src/structures/game_session.ts
@@ -3,6 +3,7 @@ import * as uuid from "uuid";
 import Eris from "eris";
 import _ from "lodash";
 
+import { E_TIMEOUT } from "async-mutex";
 import {
     bold,
     chunkArray,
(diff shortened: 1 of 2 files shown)


src/structures/game_session.ts
```diff
@@ -221,9 +222,30 @@ export default class GameSession extends Session {
...
+                        `gid: ${this.guildID} | startRound() waited ${waitMs}ms for lifecycleMutex`,
+                    );
+                }
+
+                return this.startRoundCore(messageContext);
+            });
+        } catch (e) {
+            if (e === E_TIMEOUT) {
+                logger.error(
+                    `gid: ${this.guildID} | DEADLOCK: startRound() could not acquire lifecycleMutex after 30s — force-removing session`,
```
taahamahdi, line 240: Nit: We wait 5s but this mentions 30s

taahamahdi (approved):

src/structures/game_session.ts
Brainicism: the global mutex timeout is 30s, which would throw this exception. the 5s log is for when we take >5s to obtain the lock

PR #2348, opened 2026-05-02
Fix better_audio_link race condition causing all values to reset to NULL

Brainicism:
When GenerateExpectedAvailableSongs runs concurrently (two seed processes overlapping), the second call's UPDATE LEFT JOIN finds no match for id_better_audio rows already deleted by the first call's DELETE step, silently overwriting better_audio_link with NULL for every song.

Use CASE to distinguish three states: id_better_audio is NULL (clear to NULL), the referenced row exists (use its vlink), or the row is already gone (preserve the existing column value set by the first call).


taahamahdi (approved):

Brainicism:
@copilot resolve the merge conflicts in this pull request

PR #2350, opened 2026-05-02
refactor: save better_audio files by actual YouTube ID

Brainicism:
## Problem

Songs with a `better_audio_link` were downloaded from the better audio source but saved under the **original** video's YouTube ID. This hid which video the audio actually came from — if better_audio download silently failed and fell back, or if the mapping changed, the filename told you nothing about what was really inside the `.ogg` file.

## Solution

Save files under the ID of the video that was **actually downloaded**. At playback time, resolve the correct file with a fallback chain.

## Changes (6 files)

| File | Change |
|------|--------|
| `kmq_song_downloader.ts` | `downloadYouTubeAudioWithFallback` returns the downloaded ID, files saved as `{downloadedId}.m4a/.ogg`, is-downloaded + not_downloaded + songOverrides checks match either ID |
| `session.ts` | New `resolveSongAudioPath()` resolves better_audio file on disk with fallback to original, `cached_song_duration` lookup uses `audioFileId` |
| `seed_db.ts` | `checkModifiedBetterAudioLinks` deletes/renames files by the old better_audio ID |
| `lookup.ts` | Duration lookup resolves via `betterAudioLink` — users see correct duration without any awareness of better_audio |
| `remove-dangling-songs.ts` | Collects both `link` and `better_audio_link` as valid filenames to prevent false deletions |
| `procedure 030` | `cached_song_duration` JOIN uses `COALESCE(better_audio_link, link)` |

## Deployment note

Existing songs on disk are named by the original ID. After deploying, either:
1. Run a one-time migration script to rename files, or
2. Let the next full re-download align filenames naturally

New downloads will use the correct naming immediately. The playback fallback in `resolveSongAudioPath()` handles the transition period gracefully — it checks for the better_audio file first, falls back to the original filename if missing.

### DIFF
--- a/src/helpers/kmq_song_downloader.ts
+++ b/src/helpers/kmq_song_downloader.ts
@@ -98,8 +98,11 @@ export default class KmqSongDownloader {
             let songsToDownload = allSongs;
 
             if (songOverrides) {
-                songsToDownload = songsToDownload.filter((x) =>
-                    songOverrides.includes(x.youtubeLink),
+                songsToDownload = songsToDownload.filter(
+                    (x) =>
+                        songOverrides.includes(x.youtubeLink) ||
+                        (x.betterAudioLink !== null &&
+                            songOverrides.includes(x.betterAudioLink)),
                 );
             }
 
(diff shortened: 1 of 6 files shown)


src/helpers/kmq_song_downloader.ts
```diff
@@ -240,13 +248,19 @@ export default class KmqSongDownloader {
...
                     logger.info(
-                        `Encoding song: '${song.songName}' by ${song.artistName} | ${song.youtubeLink}`,
+                        `Encoding song: '${song.songName}' by ${song.artistName} | ${downloadedId}`,
                     );
                     try {
-                        await this.encodeToOpus(cachedSongLocation, db);
+                        await this.encodeToOpus(
+                            path.join(
+                                process.env.SONG_DOWNLOAD_DIR as string,
+                                `${downloadedId}.m4a`,
```
taahamahdi, line 257: Is it expected to output as m4a?

PR #2398, opened 2026-07-15
Add player chat with a lenient profanity filter

taahamahdi:
Players can talk to each other via a collapsible chat panel, in web rooms and the embedded Activity alike.

## How it works
Modeled on the existing emote broadcast path: `POST /api/activity/chat` → hub → worker, which rate-limits per user, filters the message, and rebroadcasts a `chat` event over the existing game websocket. Messages are ephemeral (no backlog on join).

Authorization is handled at the HTTP edge — `requireAuthedInstance` already verifies the sender is a participant of the instance/room — so the worker only resolves the display identity: the web-room roster for guests, the Discord user cache for Activity users. That also lets chat work in the lobby between games.

## The filter
Deliberately lenient — common profanity is left alone.

Per @DannyKong12's suggestion, both the vocabulary **and its categorisation** come from packages rather than being written into this repo:

- [`@dsojevic/profanity-list`](https://www.npmjs.com/package/@dsojevic/profanity-list) supplies the terms, each tagged with a category and a severity, so the policy here is a per-category severity ceiling rather than any word list.
- [`obscenity`](https://www.npmjs.com/package/obscenity) does the matching, handling character substitution and the innocent-substring false positives a naive matcher trips on.

Three implementation notes worth review:

- **Severity is not a global axis.** The dataset's own README notes most entries were given a default level before being classified. Its categories are not comparable to each other, so the ceiling is set per category — a single global threshold would filter ordinary swearing while permitting things it shouldn't.
- **Only the character-level transformers are used.** obscenity's recommended bundle also collapses repeated characters, which rewrites the text being checked and therefore expects patterns pre-collapsed — correct for its own hand-written patterns, wrong for ours, which come from data and keep their natural spelling. Patterns are instead run through the same character transformers as the input so the two line up.
- **Exception expansion is guarded.** An entry's exceptions can expand into another spelling of that same entry, which would whitelist the term itself; those expansions are skipped.

## Verification
- root `tsc` 0 errors · `lint-ci` · `lint-i18n-ci` · prettier: clean
- Activity `tsc` + `vite build`: clean
- `chat_profanity_filter.test.ts`: 12 passing
- Validated locally against real inputs (not committed): every filtered spelling masks, 0 of 290 permitted terms wrongly masked, and ordinary chat passes through untouched

---

🤖 This very easy change was messed up multiple times by Claude Code


DannyKong12:
I'm not sure this is entirely the right way to go @taahamahdi, maybe use a package like https://www.npmjs.com/package/obscenity

Brainicism:
> I'm not sure this is entirely the right way to go @taahamahdi, maybe use a package like https://www.npmjs.com/package/obscenity

We only use organic homegrown profanity filters at KMQ corp

